PR #954, opened 2023-06-09
Feature/polkadot vault

valentunn:
#85zt81339

image
image
image
image
image
image
image
image
image


app/src/main/java/io/novafoundation/nova/app/root/navigation/Navigator.kt
```diff
@@ -590,7 +595,11 @@ class Navigator(
     }
 
     override fun openStartImportParitySigner() {
-        navController?.navigate(R.id.action_welcomeFragment_to_import_parity_signer_graph)
+        openStartImportPolkadotVault(PolkadotVaultVariant.PARITY_SIGNER)
```
antonijzelinskij, line 598: Am I right that we can't use Parity Signer for eth network?

antonijzelinskij (approved):

app/src/main/java/io/novafoundation/nova/app/root/navigation/Navigator.kt
valentunn: Yes, Parity Signer does not support derivation schemes used in EVM networks. Polkadot Vault does support them but we need to implement it separately (not now)

PR #967, opened 2023-06-16
Fixed transaction history reset

antonijzelinskij:
#860qx5r68


valentunn (approved):

feature-assets/src/main/java/io/novafoundation/nova/feature_assets/presentation/balance/detail/BalanceDetailFragment.kt
```diff
@@ -105,8 +105,6 @@ class BalanceDetailFragment : BaseFragment<BalanceDetailViewModel>() {
     }
 
     override fun subscribe(viewModel: BalanceDetailViewModel) {
-        viewModel.sync()
```
valentunn, line 108: Why was it causing the mentioned bug?

antonijzelinskij: Fragment calls onViewCreated after back navigation so subscribe is called every time we go back from transaction details

PR #975, opened 2023-06-21
Caching reward periods separatelly

antonijzelinskij:
#860qwu08m
#864et63rf


core-db/src/main/java/io/novafoundation/nova/core_db/migrations/42_43_StakingRewardPeriods.kt
```diff
@@ -0,0 +1,28 @@
...
+            """
+            CREATE TABLE IF NOT EXISTS `staking_reward_period` (
+            `accountId` BLOB NOT NULL, 
+            `chainId` TEXT NOT NULL, 
+            `assetId` INTEGER NOT NULL, 
+            `stakingType` TEXT NOT NULL, 
+            `periodType` TEXT NOT NULL, 
+            `customPeriodStart` INTEGER, 
+            `customPeriodEnd` INTEGER, 
+            `offsetFromCurrentDate` INTEGER,
```
ERussel, line 23: Why do we need to store offset from current date?

ERussel (approved):

core-db/src/main/java/io/novafoundation/nova/core_db/migrations/42_43_StakingRewardPeriods.kt
antonijzelinskij: I use it to save period in time (week, month, quarter etc.)

PR #1008, opened 2023-07-13
Fixed amount formating

antonijzelinskij:
#85ztgq1b5

### DIFF
--- a/common/src/main/java/io/novafoundation/nova/common/utils/formatting/DynamicPrecisionFormatter.kt
+++ b/common/src/main/java/io/novafoundation/nova/common/utils/formatting/DynamicPrecisionFormatter.kt
@@ -6,6 +6,7 @@ import java.math.RoundingMode
 import kotlin.math.min
 
 class DynamicPrecisionFormatter(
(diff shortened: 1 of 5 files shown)


common/src/main/java/io/novafoundation/nova/common/utils/formatting/DynamicPrecisionFormatter.kt
```diff
@@ -11,7 +12,8 @@ class DynamicPrecisionFormatter(
     override fun format(number: BigDecimal, roundingMode: RoundingMode): String {
         // scale() - total amount of digits after 0.,
         // precision() - amount of non-zero digits in decimal part
-        val requiredPrecision = number.scale() - number.precision() + 1
+        val zeroPrecision = number.scale() - number.precision()
+        val requiredPrecision = zeroPrecision + min(number.precision(), 3)
```
ERussel, line 16: Let't pass offset which is now hardcoded to 3 as a parameter

ERussel (approved):

PR #1014, opened 2023-07-17
Update NumberFormatters.kt

antonijzelinskij:


common/src/main/java/io/novafoundation/nova/common/utils/formatting/NumberFormatters.kt
```diff
@@ -27,7 +27,8 @@ private const val GROUPING_SEPARATOR = ','
 private const val DECIMAL_SEPARATOR = '.'
 
 private const val MIN_SCALE = 5
-private const val MIN_PRECISION = 3
+private const val AMOUNT_MIN_PRECISION = 3
```
ERussel, line 30: I suggest to use "PRICE_MIN_PRECISION"

ERussel (approved):

PR #1017, opened 2023-07-18
Update github links

valentunn:
#85ztgg3tz


ERussel (commented):
We need to update all GitHub links in the app from nova-wallet organisation to novasamatech

ERussel (approved):

PR #1019, opened 2023-07-18
Phishing protection - hide only erc20 zero transfers

valentunn:
#85ztb5cw5


feature-wallet-api/src/main/java/io/novafoundation/nova/feature_wallet_api/domain/model/Operation.kt
```diff
@@ -66,6 +67,10 @@ fun Operation.Type.satisfies(filters: Set<TransactionFilter>): Boolean {
     return matchingTransactionFilter() in filters
 }
 
+fun Operation.isZeroTransfer(): Boolean {
+    return type is Operation.Type.Transfer && type.amount.isZero
```
ERussel, line 71: Do we also need to check that this is ERC 20?

ERussel (approved):

PR #1020, opened 2023-07-18
Fix - mismatch between presets and same custom period

valentunn:
Difference was casued by the fact that preset uses date with time as a starting point whereas custom period uses beginning of a day
Also performed minor refactoring to improve type safety of `RewardPeriodType`

#85ztj1nqr


feature-staking-impl/src/main/java/io/novafoundation/nova/feature_staking_impl/domain/period/RewardPeriod.kt
```diff
@@ -1,63 +1,82 @@
...
     data class OffsetFromCurrent(
-        val offsetMillis: Long,
-        override val type: RewardPeriodType
+        val offset: Duration,
+        override val type: RewardPeriodType.Preset
     ) : RewardPeriod {
-        override val start: Date? = null
+
+        override val start: Date
+            get() = Date(System.currentTimeMillis() - offset.inWholeMilliseconds).atTheBeginningOfTheDay()
```
valentunn, line 23: This is the fix

PR #1038, opened 2023-07-26
Nomination pools/optimize era calculator

valentunn:
Greatly optimizes how data required for EraTimeCalculator is loaded. Previosuly each `factory.create` performed ~5 network requests. In worst case, we had to create this factory twice - one for waiting stake summary for unbondings. That caused signifiacnt delay before showing unbondings

This PR does several optimizations:

1. We now cache currentSlot, genesisSlot, eraStartIndex, currentSessionIndex,  currentEpochIndex
2. EraTimeCalculator is now shared via StakingSharedComputation

To mitigate outdated values in cache, they are observed

### DIFF
--- a/feature-staking-impl/src/main/java/io/novafoundation/nova/feature_staking_impl/data/network/blockhain/api/SessionRuntimeApi.kt
+++ b/feature-staking-impl/src/main/java/io/novafoundation/nova/feature_staking_impl/data/network/blockhain/api/SessionRuntimeApi.kt
@@ -1,0 +1,23 @@
+package io.novafoundation.nova.feature_staking_impl.data.network.blockhain.api
+
+import io.novafoundation.nova.common.utils.session
+import io.novafoundation.nova.feature_staking_impl.data.network.blockhain.bindings.bindSessionIndex
+import io.novafoundation.nova.runtime.storage.source.query.StorageQueryContext
+import io.novafoundation.nova.runtime.storage.source.query.api.QueryableModule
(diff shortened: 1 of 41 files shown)


feature-staking-impl/src/main/java/io/novafoundation/nova/feature_staking_impl/data/network/blockhain/api/SessionRuntimeApi.kt
```diff
@@ -0,0 +1,22 @@
...
+@JvmInline
+value class SessionRuntimeApi(override val module: Module) : QueryableModule
+
+context(StorageQueryContext)
+val RuntimeMetadata.session: SessionRuntimeApi
+    get() = SessionRuntimeApi(session())
+
+context(StorageQueryContext)
+val SessionRuntimeApi.currentIndex: QueryableStorageEntry0<BigInteger>
+    get() = storage0("CurrentIndex", binding = ::bindSessionIndex)
```
antonijzelinskij, line 22: What does the number in the `storage` fun name mean?

antonijzelinskij (approved):

feature-staking-impl/src/main/java/io/novafoundation/nova/feature_staking_impl/data/network/blockhain/api/SessionRuntimeApi.kt
valentunn: Its number of arguments, 
so storage0 has no arguments
storage1 has one

PR #1054, opened 2023-08-09
move changelog to the main repo

stepanLav:


valentunn (approved):
Looks like we need to create a task to change urls in app accordingly

PR #1061, opened 2023-08-14
Feature/fee spikes

valentunn:
Intoduce `Fee` interface which abstracts different kinds of fee structures, e.g. `evm(gasLimit, gasPrice)`, `other(particalFee)`

Refactor `FeeLoaderMixin` to operate with `Fee` instead of plan `Balance`. Introduce backward-compatible adapter methods to avoid changes in screens that use `FeeLoaderMixin`

Introduce `FeeChangeValidation` and corresponding failure handler that overwrites fee in `FeeLoaderMixin`. If fee difference is too much (ratio is grater than 1.5) show confirmation warning before modifying the fee. For now, this validation is only applied to EVM logic - EvmNative, EvmErc20 send, EvmExternalSign.

Let `EvmTransactionService` to accept optional `presetFee` parameter to forcefully use that as a fee if provided

### Warning examples (lowered threshold)
image
image


feature-account-impl/src/main/java/io/novafoundation/nova/feature_account_impl/data/extrinsic/RealExtrinsicService.kt
```diff
@@ -118,10 +120,14 @@ class RealExtrinsicService(
         extrinsicBuilder.formExtrinsic()
         val extrinsic = extrinsicBuilder.build()
 
-        return estimateFee(chain.id, extrinsic)
+        return estimateFee(chain.id, extrinsic).amount
     }
 
-    override suspend fun estimateFee(chainId: ChainId, extrinsic: String): BigInteger {
+    override suspend fun estimateFeeV2(chain: Chain, formExtrinsic: suspend ExtrinsicBuilder.() -> Unit): Fee {
```
antonijzelinskij, line 126: can we rename this function to estimateFee to overload it?

antonijzelinskij (approved):

feature-account-impl/src/main/java/io/novafoundation/nova/feature_account_impl/data/extrinsic/RealExtrinsicService.kt
valentunn: We cannot. they will only differ by return type, which is not included into method signature so they will be considered same from overload resolution perspective

PR #1082, opened 2023-08-28
Remove unused code and logic

valentunn:
* Remove start staking widget and related components
* Remove setup staking screen
* Remove bond logic from `SetupStakingState`
* Remove bond logic from `ConfirmStaking` screen, rename it to `ConfirmChangeValidatorsViewModel` screen


feature-staking-impl/src/main/java/io/novafoundation/nova/feature_staking_impl/di/StakingFeatureModule.kt
```diff
@@ -319,7 +319,7 @@ class StakingFeatureModule {
     fun provideSetupStakingInteractor(
         extrinsicService: ExtrinsicService,
         sharedState: StakingSharedState,
-    ) = SetupStakingInteractor(extrinsicService, sharedState)
+    ) = ChangeValidatorsInteractor(extrinsicService, sharedState)
```
antonijzelinskij, line 322: It seems we also should change fun name

antonijzelinskij (approved):

PR #1093, opened 2023-09-06
Use default block time from chains config

valentunn:
#85ztrenxh


stepanLav:
Task linked: [CU-85ztrenxh Android. Use default blocktime for chains from chains.json](https://app.clickup.com/t/85ztrenxh)

common/src/main/java/io/novafoundation/nova/common/utils/JsonExt.kt
```diff
@@ -13,6 +13,15 @@ fun Any?.asGsonParsedNumberOrNull(): BigInteger? = when (this) {
     else -> null
 }
 
+fun Any?.asGsonParsedLongOrNull(): Long? = when (this) {
+    // gson parses integers as double when type is not specified
+    is Double -> toLong()
+    is Long -> this
+    is Int -> toLong()
```
antonijzelinskij, line 20: We also can use Number instead of Double, Long, Int

antonijzelinskij (approved):

PR #1108, opened 2023-09-11
Fix/open keyboard setup amount

antonijzelinskij:


common/src/main/res/values/strings.xml
```diff
@@ -1,6 +1,8 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <resources xmlns:tools="http://schemas.android.com/tools" tools:ignore="TypographyEllipsis">
 
+    <string name="transactions_pool_staking_subtitle">Pool staking</string>
```
valentunn, line 4: I suggest reusing `setup_staking_type_pool_staking`

valentunn (approved):

PR #1119, opened 2023-09-14
Fix/tx history updates

valentunn:
#85ztrxnyf

Status was not updating becouse of `PayloadDiffer` finding difference for `amountWithDetails` and thus updating `amountWithDetails` in `resolvePayload`, ignoring updates of `statusAppearance`

### DIFF
--- a/feature-assets/src/main/java/io/novafoundation/nova/feature_assets/presentation/transaction/history/TransactionHistoryAdapter.kt
+++ b/feature-assets/src/main/java/io/novafoundation/nova/feature_assets/presentation/transaction/history/TransactionHistoryAdapter.kt
@@ -97,7 +97,7 @@ class TransactionHolder(
     }
 
     fun bindHeader(item: OperationModel) {
-        containerView. header.text = item.header
+        containerView.header.text = item.header
     }
 
     fun bindAmountDetails(item: OperationModel) {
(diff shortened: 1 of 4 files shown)


stepanLav:
Task linked: [CU-85ztrxnyf Android. Fix EVM transactions status realtime update](https://app.clickup.com/t/85ztrxnyf)

feature-assets/src/main/java/io/novafoundation/nova/feature_assets/presentation/transaction/history/TransactionHistoryAdapter.kt
```diff
@@ -70,32 +76,49 @@ class TransactionHolder(
...
-                status.makeVisible()
-                status.setImageResource(item.statusAppearance.icon)
-                status.setImageTintRes(item.statusAppearance.statusIconTint)
-            } else {
-                status.makeGone()
-            }
+            bindStatus(item)
 
-            setOnClickListener { handler.transactionClicked(item) }
+            setOnClickListener { handler.transactionClicked(item.id) }
```
valentunn, line 90: This is to avoid potentially outdated model arriving to `handler.transactionClicked` since we do not re-set onClickListener. We only use `item.id` anyway

antonijzelinskij (approved):

PR #1131, opened 2023-09-19
Fixed open pin code twice

antonijzelinskij:
Test case:
Precondition: enabled "don't keep activities"
- open app on main screen
- hide app to let system kill it
- open app (the pin verification must open)
- hide app to let system kill the app again
- open app

Result: Second pin verification fragment is shown. To use the app we must input pin code twice on both screens


antonijzelinskij (commented):
This fix doesn't consider that we also should open pin verification on changePinFlow

valentunn:
> This fix doesn't consider that we also should open pin verification on changePinFlow

You mean this PR is still WIP?

PR #1133, opened 2023-09-21
Filtered active nominators in min stake calculation

antonijzelinskij:
#85ztn660y

### DIFF
--- a/feature-staking-impl/src/main/java/io/novafoundation/nova/feature_staking_impl/domain/StakingInteractorExt.kt
+++ b/feature-staking-impl/src/main/java/io/novafoundation/nova/feature_staking_impl/domain/StakingInteractorExt.kt
@@ -63,6 +63,7 @@ fun minimumStake(
     exposures: Collection<Exposure>,
     minimumNominatorBond: BigInteger,
     bagListLocator: BagListLocator?,
+    maxNominatorsInValidator: Int,
     bagListScoreConverter: BagListScoreConverter,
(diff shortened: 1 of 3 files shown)


stepanLav:
Task linked: [CU-85ztn660y Difference in minStake for Westend](https://app.clickup.com/t/85ztn660y)

feature-staking-impl/src/main/java/io/novafoundation/nova/feature_staking_impl/domain/StakingInteractorExt.kt
```diff
@@ -71,11 +72,13 @@ fun minimumStake(
 
     val stakeByNominator = exposures
         .fold(mutableMapOf<AccountIdKey, BigInteger>()) { acc, exposure ->
-            exposure.others.forEach { individualExposure ->
-                val key = individualExposure.who.intoKey()
-                val currentExposure = acc.getOrDefault(key, BigInteger.ZERO)
-                acc[key] = currentExposure + individualExposure.value
-            }
+            exposure.others
+                .take(maxNominatorsInValidator)
```
valentunn, line 76: Shouldn't `exposure.others` be sorted before taking first `maxNominatorsInValidator`?

antonijzelinskij: We get nominators with nice sort from on chain so I think it's not necessary

PR #1134, opened 2023-09-21
Add existential deposit validation

antonijzelinskij:
#8692t0fq5

### DIFF
--- a/feature-staking-impl/src/main/java/io/novafoundation/nova/feature_staking_impl/domain/nominationPools/claimRewards/validations/Declarations.kt
+++ b/feature-staking-impl/src/main/java/io/novafoundation/nova/feature_staking_impl/domain/nominationPools/claimRewards/validations/Declarations.kt
@@ -3,11 +3,10 @@ package io.novafoundation.nova.feature_staking_impl.domain.nominationPools.claim
 import io.novafoundation.nova.common.validation.ValidationSystem
 import io.novafoundation.nova.common.validation.ValidationSystemBuilder
 import io.novafoundation.nova.feature_staking_impl.domain.common.validation.profitableAction
-import io.novafoundation.nova.feature_wallet_api.data.network.blockhain.assets.AssetSourceRegistry
-import io.novafoundation.nova.feature_wallet_api.data.network.blockhain.assets.existentialDeposit
-import io.novafoundation.nova.feature_wallet_api.domain.validation.enoughTotalToStayAboveED
+import io.novafoundation.nova.feature_wallet_api.domain.validation.EnoughTotalToStayAboveEDValidationFactory
 import io.novafoundation.nova.feature_wallet_api.domain.validation.sufficientBalance
 import io.novafoundation.nova.runtime.ext.utilityAsset
+import io.novafoundation.nova.runtime.multiNetwork.ChainWithAsset
 
 typealias NominationPoolsClaimRewardsValidationSystem =
     ValidationSystem<NominationPoolsClaimRewardsValidationPayload, NominationPoolsClaimRewardsValidationFailure>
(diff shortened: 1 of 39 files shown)


stepanLav:
Task linked: [CU-8692t0fq5 Android. Add eD validation for nominal pool flows](https://app.clickup.com/t/8692t0fq5)

feature-staking-impl/src/main/java/io/novafoundation/nova/feature_staking_impl/domain/nominationPools/claimRewards/validations/ValidationFailureUi.kt
```diff
@@ -15,5 +15,8 @@ fun nominationPoolsClaimRewardsValidationFailure(
 
         NominationPoolsClaimRewardsValidationFailure.NonProfitableClaim -> resourceManager.getString(R.string.common_confirmation_title) to
             resourceManager.getString(R.string.staking_warning_tiny_payout)
+
+        is NominationPoolsClaimRewardsValidationFailure.ToStayAboveED -> resourceManager.getString(R.string.common_not_enough_funds_title) to
+            resourceManager.getString(R.string.wallet_send_insufficient_balance_commission, failure.asset.symbol)
```
ERussel, line 20: Can we make a shared handler to not to duplicate title and message in each case (unstake, redeem, claim)?

antonijzelinskij: As far as I know we do not adhere to this practice for validation since we have different Failure classes for each validation

valentunn: We can actually move this to a separate function by using a separate interface for validation error. We already use it in separate placed, for example NotEnoughToPayFees validation

feature-staking-impl/src/main/java/io/novafoundation/nova/feature_staking_impl/domain/nominationPools/claimRewards/validations/Declarations.kt
```diff
@@ -3,19 +3,39 @@ package io.novafoundation.nova.feature_staking_impl.domain.nominationPools.claim
...
 
 typealias NominationPoolsClaimRewardsValidationSystem =
     ValidationSystem<NominationPoolsClaimRewardsValidationPayload, NominationPoolsClaimRewardsValidationFailure>
+
 typealias NominationPoolsClaimRewardsValidationSystemBuilder =
     ValidationSystemBuilder<NominationPoolsClaimRewardsValidationPayload, NominationPoolsClaimRewardsValidationFailure>
 
-fun ValidationSystem.Companion.nominationPoolsClaimRewards(): NominationPoolsClaimRewardsValidationSystem = ValidationSystem {
+fun ValidationSystem.Companion.nominationPoolsClaimRewards(
+    assetSourceRegistry: AssetSourceRegistry
```
valentunn, line 19: I suggest to create a factory for this validation that later can be passed to concrete validation systems. It would prevent us to modify all places where validation is used in case it will require new dependencies in the future

feature-staking-impl/src/main/java/io/novafoundation/nova/feature_staking_impl/domain/nominationPools/claimRewards/validations/Declarations.kt
```diff
@@ -3,19 +3,39 @@ package io.novafoundation.nova.feature_staking_impl.domain.nominationPools.claim
...
     profitableClaim()
 }
 
+private fun NominationPoolsClaimRewardsValidationSystemBuilder.sufficientCommissionBalanceToStayAboveED(
+    assetSourceRegistry: AssetSourceRegistry
+) {
+    enoughTotalToStayAboveED(
+        fee = { it.fee },
+        total = { it.asset.total },
+        existentialDeposit = { assetSourceRegistry.existentialDeposit(it.chain, it.chain.utilityAsset) },
```
valentunn, line 34: Seems like this is the common part so we can probably pass asset registry directly and use it there

feature-staking-impl/src/main/java/io/novafoundation/nova/feature_staking_impl/domain/nominationPools/claimRewards/validations/Declarations.kt
```diff
@@ -16,23 +15,25 @@ typealias NominationPoolsClaimRewardsValidationSystemBuilder =
...
-        total = { it.asset.total },
-        existentialDeposit = { assetSourceRegistry.existentialDeposit(it.chain, it.chain.utilityAsset) },
-        error = { NominationPoolsClaimRewardsValidationFailure.ToStayAboveED(it.chain.utilityAsset) }
+    validate(
+        enoughTotalToStayAboveEDValidationFactory.create(
+            fee = { it.fee },
+            total = { it.asset.total },
+            chainWithAsset = { ChainWithAsset(it.chain, it.chain.utilityAsset) },
+            error = { NominationPoolsClaimRewardsValidationFailure.ToStayAboveED(it.chain.utilityAsset) }
+        )
```
valentunn, line 36: We can remove the need to call `validate` every time if we call it and use `context(ValidationSystemBuilder<P, S>)` inside `enoughTotalToStayAboveEDValidationFactory`

feature-staking-impl/src/main/java/io/novafoundation/nova/feature_staking_impl/domain/nominationPools/redeem/validations/ValidationFailureUi.kt
```diff
@@ -2,17 +2,19 @@ package io.novafoundation.nova.feature_staking_impl.domain.nominationPools.redee
...
 fun nominationPoolsRedeemValidationFailure(
     failure: NominationPoolsRedeemValidationFailure,
     resourceManager: ResourceManager
 ): TitleAndMessage {
     return when (failure) {
         is NotEnoughBalanceToPayFees -> handleNotEnoughFeeError(failure, resourceManager)
-        is NominationPoolsRedeemValidationFailure.ToStayAboveED -> resourceManager.getString(R.string.common_not_enough_funds_title) to
-            resourceManager.getString(R.string.wallet_send_insufficient_balance_commission, failure.asset.symbol)
+        is NominationPoolsRedeemValidationFailure.ToStayAboveED -> handleInsufficientBalanceCommission(
+            failure.asset,
```
valentunn, line 16: I suggest to pass `failure` directly to `handleInsufficientBalanceCommission`. For that, we can create an interface `InsufficientTotalToStayAboveEDError` which contains `all fields necessary for displaying an error. This will be useful in case we want to change display logic and add/remove extra information

antonijzelinskij: I think it's not really easy to provide this fix since we use different failure classes

valentunn (approved):

PR #1136, opened 2023-09-22
Fixed crash after change wallet

antonijzelinskij:
#8692t85a6


stepanLav:
Task linked: [CU-8692t85a6 Android. Crash after switching accounts and go to Send](https://app.clickup.com/t/8692t85a6)

valentunn (approved):
Can you explain the cause of the crash?

antonijzelinskij:
> Can you explain the cause of the crash?

I'm not sure because didn't catch it on android. But the logs in the task tells that the crash happens since RecyclerView is null during call

valentunn:
> > Can you explain the cause of the crash?
> 
> I'm not sure because didn't catch it on android. But the logs in the task tells that the crash happens since RecyclerView is null during call

In that case `observeWhenCreated` wont actually change much since `observe` we're using is an extension for `launchWhenResumed` which happens even later in lifecycle

PR #1140, opened 2023-09-25
Sort asset groups by transferable fiat balance

antonijzelinskij:
#8692t8mcn


stepanLav:
Task linked: [CU-8692t8mcn Android. Apply sort tokens by fiat value](https://app.clickup.com/t/8692t8mcn)

feature-assets/src/main/java/io/novafoundation/nova/feature_assets/domain/assets/search/AssetSearchInteractor.kt
```diff
@@ -39,7 +40,9 @@ class AssetSearchInteractor(
         queryFlow: Flow<String>,
         externalBalancesFlow: Flow<List<ExternalBalance>>,
     ): Flow<Map<AssetGroup, List<AssetWithOffChainBalance>>> {
-        return searchAssetsInternalFlow(queryFlow, externalBalancesFlow) { asset ->
+        val comparator = compareByDescending(AssetGroup::groupTransferableBalanceFiat)
```
valentunn, line 43: We should also apply default chain sorting as a fallback `.then(Chain.defaultComparatorFrom(AssetGroup::chain))
`

PR #1144, opened 2023-09-26
Hide keyboard only for back pressing and asset selecting

antonijzelinskij:
#8692t8e34

### DIFF
--- a/feature-assets/src/main/java/io/novafoundation/nova/feature_assets/presentation/flow/AssetFlowFragment.kt
+++ b/feature-assets/src/main/java/io/novafoundation/nova/feature_assets/presentation/flow/AssetFlowFragment.kt
@@ -95,13 +95,15 @@ abstract class AssetFlowFragment<T : AssetFlowViewModel> :
         }
     }
 
-    override fun onDestroyView() {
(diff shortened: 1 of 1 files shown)


stepanLav:
Task linked: [CU-8692t8e34 Android. Enable keyboard on Buy from call to action](https://app.clickup.com/t/8692t8e34)

feature-assets/src/main/java/io/novafoundation/nova/feature_assets/presentation/flow/AssetFlowFragment.kt
```diff
@@ -95,13 +95,15 @@ abstract class AssetFlowFragment<T : AssetFlowViewModel> :
         }
     }
 
-    override fun onDestroyView() {
-        super.onDestroyView()
+    override fun onBackPressed(action: () -> Unit) {
+        super.onBackPressed(action)
```
valentunn, line 99: This is a wrong override I believe. `onBackPressed(action: () -> Unit)` is a method in BaseFragment to register a back-pressed listener. It is not a callback itself

valentunn (approved):

PR #1147, opened 2023-09-26
Swap/new modules

valentunn:


feature-swap-impl/build.gradle
```diff
@@ -0,0 +1,69 @@
...
+
+dependencies {
+    implementation project(':feature-account-api')
+    implementation project(':feature-wallet-api')
+    implementation project(':feature-swap-api')
+
+    implementation project(":common")
+    implementation project(":runtime")
+
+    implementation markwonDep
```
antonijzelinskij, line 44: Do we really need markdown dependency here?

antonijzelinskij (approved):

feature-swap-impl/build.gradle
valentunn: Nice catch, it was a copy-paste from gov module, I will remove this dependency

PR #1171, opened 2023-10-16
Feature/swap main logic

valentunn:


common/src/main/java/io/novafoundation/nova/common/utils/UserEditableString.kt
```diff
@@ -0,0 +1,3 @@
+package io.novafoundation.nova.common.utils
+
+data class UserEditableString(val value: String = "", val editedByUser: Boolean = false)
```
antonijzelinskij, line 3: I guess we can remove it

antonijzelinskij (approved):

PR #1177, opened 2023-10-20
Swap options

antonijzelinskij:
#8692ty65z


stepanLav:
Task linked: [CU-8692ty65z Android. Slippage settings screen](https://app.clickup.com/t/8692ty65z)

stepanLav:
Task linked: [CU-8692ty65z Android. Slippage settings screen](https://app.clickup.com/t/8692ty65z)

stepanLav:
Task linked: [CU-8692ty65z Android. Slippage settings screen](https://app.clickup.com/t/8692ty65z)

common/src/main/java/io/novafoundation/nova/common/utils/FlowExt.kt
```diff
@@ -308,6 +308,9 @@ inline fun <T> EditText.bindTo(
             if (text.toString() != inputString) {
                 removeTextChangedListener(textWatcher)
                 setText(inputString)
+                if (hasFocus()) {
+                    setSelection(inputString.length)
+                }
```
valentunn, line 313: What if user is editing field in the middle?

common/src/main/java/io/novafoundation/nova/common/view/TipsInputField.kt
```diff
@@ -0,0 +1,171 @@
...
+
+        content.onTextChanged {
+            tipsInputClear.isVisible = it.isNotEmpty()
+            tipsInputContainer.isVisible = it.isEmpty()
+            measurePostfix(it)
+            invalidate()
+        }
+
+        textPaint = Paint(content.paint).apply {
+            color = content.currentTextColor
```
valentunn, line 60: Doesn't `Paint(content.paint)` also copy the color?

common/src/main/java/io/novafoundation/nova/common/validation/InputValidationMixinFactory.kt
```diff
@@ -0,0 +1,25 @@
...
+) {
+    fun create(inputFlow: Flow<String>, payloadFormatter: (String) -> P): InputValidationMixin<P, E> {
+        return InputValidationMixin(validationSystem, inputFlow, payloadFormatter)
+    }
+}
+
+class InputValidationMixin<P, E>(
+    private val validationSystem: ValidationSystem<P, E>,
+    private val inputFlow: Flow<String>,
+    private val payloadFormatter: (String) -> P
```
valentunn, line 17: We can pass `Flow<P>` to be more generic and let the factory handle the transformation

common/src/main/java/io/novafoundation/nova/common/view/TipsInputField.kt
```diff
@@ -0,0 +1,171 @@
...
+import io.novafoundation.nova.common.utils.useAttributes
+import io.novafoundation.nova.common.view.shape.getInputBackground
+import io.novafoundation.nova.common.view.shape.getInputBackgroundError
+import kotlinx.android.synthetic.main.view_tips_input.view.tipsInputClear
+import kotlinx.android.synthetic.main.view_tips_input.view.tipsInputContainer
+import kotlinx.android.synthetic.main.view_tips_input.view.tipsInputError
+import kotlinx.android.synthetic.main.view_tips_input.view.tipsInputField
+import kotlinx.android.synthetic.main.view_tips_input.view.tipsInputFieldContainer
+
+class TipsInputField @JvmOverloads constructor(
```
valentunn, line 31: You mentioned you were also trying the approach with Matieral input. can you quickly describe why you decided to step away from that approach? Did library updating cause issues?

common/src/main/res/drawable/ic_new_info_icon_16.xml
```diff
@@ -0,0 +1,19 @@
...
+        android:pathData="M8,8m-8,0a8,8 0,1 1,16 0a8,8 0,1 1,-16 0"
+        android:fillColor="#999EC7"
+        android:fillAlpha="0.24"/>
+    <path
+        android:pathData="M7.542,5.545C7.664,5.643 7.828,5.699 8,5.701C8.171,5.699 8.335,5.643 8.457,5.545C8.578,5.447 8.647,5.314 8.65,5.176V4.826C8.65,4.687 8.581,4.553 8.459,4.455C8.337,4.356 8.172,4.301 8,4.301C7.827,4.301 7.662,4.356 7.54,4.455C7.418,4.553 7.35,4.687 7.35,4.826V5.176C7.352,5.314 7.421,5.447 7.542,5.545ZM7.542,11.519C7.664,11.628 7.828,11.69 8,11.692C8.171,11.69 8.335,11.628 8.457,11.519C8.578,11.41 8.647,11.263 8.65,11.109V7.22C8.65,7.065 8.581,6.917 8.459,6.808C8.337,6.698 8.172,6.637 8,6.637C7.827,6.637 7.662,6.698 7.54,6.808C7.418,6.917 7.35,7.065 7.35,7.22V11.109C7.352,11.263 7.421,11.41 7.542,11.519Z"
+        android:fillColor="#ffffff"
+        android:fillAlpha="0.88"
+        android:fillType="evenOdd"/>
+  </group>
+</vector>
```
valentunn, line 19: Afair we decided to actually use this info icon everwhere so seems like we can just replace old icon

feature-swap-impl/src/main/java/io/novafoundation/nova/feature_swap_impl/domain/slippage/SlippageRepository.kt
```diff
@@ -0,0 +1,45 @@
...
+
+    fun getSlippageTips(): List<Percent>
+
+    fun minSlippage(): Percent
+
+    fun maxSlippage(): Percent
+
+    fun smallSlippage(): Percent
+
+    abstract fun bigSlippage(): Percent
```
valentunn, line 17: abstract?

feature-swap-impl/src/main/java/io/novafoundation/nova/feature_swap_impl/domain/slippage/SlippageRepository.kt
```diff
@@ -0,0 +1,45 @@
...
+
+    fun getDefaultSlippage(): Percent
+
+    fun getSlippageTips(): List<Percent>
+
+    fun minSlippage(): Percent
+
+    fun maxSlippage(): Percent
+
+    fun smallSlippage(): Percent
```
valentunn, line 15: OK for now, but the target approach is to define slippage configuration inside `AssetExchange`:

``` interface AssetExchange {
...

suspend fun slippageConfig(): SlippageConfig
}
```

We can leave repository approach for now though

common/src/main/java/io/novafoundation/nova/common/validation/InputValidationMixinFactory.kt
```diff
@@ -0,0 +1,25 @@
...
+import kotlinx.coroutines.flow.Flow
+import kotlinx.coroutines.flow.map
+
+class InputValidationMixinFactory<P, E>(
+    private val validationSystem: ValidationSystem<P, E>,
+) {
+    fun create(inputFlow: Flow<String>, payloadFormatter: (String) -> P): InputValidationMixin<P, E> {
+        return InputValidationMixin(validationSystem, inputFlow, payloadFormatter)
+    }
+}
```
valentunn, line 12: To be honest I think using a validation system is overkill for field validations and it will lead to just more boilerplate.
We can come up with a simpler version that is easier to configure:
``` kotlin
interface FieldValidator {
   validate(inputStream: Flow<String>): Flow<FieldValidationResult>

  compaction object;
}

fun FieldValidator.oneOf(vararg inner: FieldValidator): FieldValidator

sealed class FieldValidationResult {
   object Ok: FieldValidationResult()

  class Error(val reason: String): FieldValidationResult()

class Warning(val reason: String): FieldValidationResult()
}
```

feature-swap-impl/src/main/java/io/novafoundation/nova/feature_swap_impl/presentation/options/SwapOptionsViewModel.kt
```diff
@@ -0,0 +1,178 @@
...
+    private val swapSettingsStateFlow = flowOfAll { swapSettingState.await().selectedOption }
+        .shareInBackground()
+
+    val slippageInput = MutableStateFlow("")
+
+    private val slippageInputValidationStatus = inputValidationMixinFactory.create(slippageInput, ::formatSlippagePayload)
+        .observeStatus()
+        .shareInBackground()
+
+    private fun formatSlippagePayload(text: String): SlippageValidationPayload {
```
valentunn, line 56: `private fun` should be at the very bottom of the class

feature-swap-impl/src/main/java/io/novafoundation/nova/feature_swap_impl/presentation/options/SwapOptionsViewModel.kt
```diff
@@ -0,0 +1,178 @@
...
+            Disabled(resourceManager.getString(R.string.swap_slippage_disabled_button_state))
+        } else {
+            val slippage = insertedSlippage.formatToPercent()
+            if (slippage != settings.slippage) {
+                Enabled(resourceManager.getString(R.string.common_apply))
+            } else {
+                Disabled(resourceManager.getString(R.string.common_apply))
+            }
+        }
+    }
```
valentunn, line 173: This could be replaced with `when`

common/src/main/java/io/novafoundation/nova/common/utils/FlowExt.kt
antonijzelinskij: I guess that nothing will happen. selection will move to end only after input flow value changed.
I think we need to have regression testing

common/src/main/java/io/novafoundation/nova/common/view/TipsInputField.kt
antonijzelinskij: Yes it does. I'll fix it

common/src/main/java/io/novafoundation/nova/common/view/TipsInputField.kt
antonijzelinskij: The postfix in Material library is placed in right side of TextInput view and there no way to put it after inserted text

common/src/main/java/io/novafoundation/nova/common/utils/FlowExt.kt
antonijzelinskij: Or we can make this logic only for this case

common/src/main/res/drawable/ic_new_info_icon_16.xml
antonijzelinskij: It looks like it would be better to do this as a separate task, because... In many places there is a filter applied to this icon and we will have to manually make some changes

common/src/main/java/io/novafoundation/nova/common/utils/FlowExt.kt
valentunn: But why is this needed in the first place? I thought this is no longer a problem for swap main screen

antonijzelinskij: I need this for slippage input field. When I click tips the text will change but selection will not move to the end

valentunn: Lets enable it only by a boolean flag passed to the function to avoid alter default behavior and avoid possible minor regressions

feature-swap-impl/src/main/java/io/novafoundation/nova/feature_swap_impl/data/assetExchange/assetConversion/AssetConversionExchange.kt
```diff
@@ -121,6 +123,17 @@ private class AssetConversionExchange(
...
     }
 
+    override suspend fun slippageConfig(): SlippageConfig {
+        return SlippageConfig(
+            Percent(0.5),
+            listOf(Percent(0.1), Percent(1.0), Percent(3.0)),
+            Percent(0.1),
+            Percent(50.0),
+            Percent(0.5),
+            Percent(2.5)
```
valentunn, line 133: Can we use named arguments here?

common/src/main/java/io/novafoundation/nova/common/validation/FieldValidator.kt
```diff
@@ -0,0 +1,17 @@
+package io.novafoundation.nova.common.validation
+
+import kotlinx.coroutines.flow.Flow
+
+interface FieldValidator {
+    fun observe(inputStream: Flow<String>): Flow<FieldValidationResult>
+
+    companion object
+}
```
valentunn, line 9: What do you think about making a small subclass named something like `MapFieldValidator`?

``` kotlin

class MapFieldValidator : FieldValidator {
  abstract fun validate(input: String): FieldValidationResult

 override fun observe(inputStream: Flow<String>) = inputStream.map(::validate)
}
```

This will reduce boilerplate for simpler validations like slippage one

feature-swap-impl/src/main/java/io/novafoundation/nova/feature_swap_impl/presentation/options/SwapOptionsViewModel.kt
```diff
@@ -10,84 +10,87 @@ import io.novafoundation.nova.common.utils.Percent
...
         swapSettingsStateProvider.getSwapSettingsState(viewModelScope)
     }
 
     private val swapSettingsStateFlow = flowOfAll { swapSettingState.await().selectedOption }
         .shareInBackground()
 
+    private val assetExchange = swapSettingsStateFlow.mapNotNull { it.assetIn }
+        .map { assetExchangeFactory.create(it.chainId) }
+        .filterNotNull()
+        .shareInBackground()
```
valentunn, line 48: `AssetExchange` interfacge shouldnt be exposed to presentationlayer, since it is in `data` layer. And it shouldnt be created via factory since lifecycle of `AssetExchange` instances is managed by `SwapService`. You can propagate config calls via `SwapService` and `SwapInteractor` by passing `chain` to retrieve relevant exchange in `swapService` similar to other methods in `SwapService`

feature-swap-impl/src/main/java/io/novafoundation/nova/feature_swap_impl/presentation/fieldValidation/SlippageFieldValidator.kt
```diff
@@ -0,0 +1,61 @@
...
+
+class SlippageFieldValidatorFactory(private val resourceManager: ResourceManager) {
+
+    suspend fun create(assetExchange: AssetExchange): SlippageFieldValidator {
+        return SlippageFieldValidator(assetExchange, resourceManager)
+    }
+}
+
+class SlippageFieldValidator(
+    private val assetExchange: AssetExchange,
```
valentunn, line 21: See comment regarding exposure of `Assetxchange` below

feature-swap-impl/src/main/java/io/novafoundation/nova/feature_swap_impl/presentation/fieldValidation/SlippageFieldValidator.kt
```diff
@@ -0,0 +1,61 @@
...
+
+    override fun observe(inputStream: Flow<String>): Flow<FieldValidationResult> {
+        return inputStream.map {
+            val slippageConfig = assetExchange.slippageConfig()
+            val value = it.toPercent()
+            when {
+                it.isEmpty() -> FieldValidationResult.Ok
+                value == null -> FieldValidationResult.Error(resourceManager.getString(R.string.common_error_general_title))
+
+                value < slippageConfig.minAvailableSlippage || value > slippageConfig.maxAvailableSlippage -> {
```
valentunn, line 33:
``` kotlin
value !in slippageConfig.minAvailableSlippage..slippageConfig.maxAvailableSlippage
```

feature-swap-impl/src/main/java/io/novafoundation/nova/feature_swap_impl/presentation/options/SwapOptionsViewModel.kt
```diff
@@ -10,84 +10,87 @@ import io.novafoundation.nova.common.utils.Percent
...
-    private val slippageInputValidationStatus = inputValidationMixinFactory.create(slippageInput, ::formatSlippagePayload)
-        .observeStatus()
+    private val slippageFieldValidator = assetExchange.map { slippageFieldValidatorFactory.create(it) }
         .shareInBackground()
 
-    private fun formatSlippagePayload(text: String): SlippageValidationPayload {
-        return SlippageValidationPayload(text.formatToPercent())
-    }
+    private val slippageInputValidationResult = slippageFieldValidator.flatMapLatest { it.observe(slippageInput) }
+        .shareInBackground()
```
valentunn, line 59: We should always think about code reusability and code dublication. Given that the very next task is to apply the same logic to multiple fields in the setup swap screen, it is really worth to make field validation easy to reuse from the beginning. Here are some thoughts on how it can potentially be achieved (feel freee to adjust):

``` kotlin
// -------------- views ------------------

interface ValidatableInputField {
  fun showError(error: String)
  
  fun hideError()
}

class SwapInputView : ValidatableInputField 
class SlippageInputView : ValidatableInputField // I dont rember the exact name for this view, idea is to implement interface for every validatrable field
 
// --------- presentation --------------

interface InputValidationMixin {
  val fieldValidationResult: Flow<FieldValidationResult>

  // this factory can be used standalone (for slippage) or inside other mixins like SwapAmountInputMixin
  interface Factory {
    fun create(coroutineScope: CoroutineScope, inputStream: Flow<String>, validator: FieldValidator)
  }
}

class SwapAmountInputMixinFactory(..) {
 fun create(
        ....
        validator: FieldValidator,
    ): SwapAmountInputMixin.Presentation {
        return RealSwapAmountInputMixin(
            ....
            validator: FieldValidator,
        )
    }
}

**Review by valentunn — CHANGES_REQUESTED** [review]

File: common/src/main/java/io/novafoundation/nova/common/utils/FlowExt.kt [hunk 17]
```diff
@@ -308,15 +315,42 @@ inline fun <T> EditText.bindTo(
...
+
+                is FieldValidationResult.Error -> {
+                    showError(validationResult.reason)
+                }
+
+                else -> {}
+            }
+        }
+    }
+}
```
valentunn, line 346: I think more appropriate place for this is somewhere neraby FieldValidator itself

common/src/main/java/io/novafoundation/nova/common/validation/InputValidationMixin.kt
```diff
@@ -0,0 +1,26 @@
...
+    interface Factory {
+        fun create(inputStream: Flow<String>, validator: FieldValidator)
+    }
+}
+
+class RealInputValidationMixinFactory(val validator: FieldValidator) {
+
+    fun create(inputStream: Flow<String>): InputValidationMixin {
+        return RealInputValidationMixin(inputStream, validator)
+    }
```
valentunn, line 17: I think it should be this way:

``` kotlin
RealInputValidationMixinFactory : InputValidationMixin.Factory {
   override fun create(inputStream: Flow<String>, validator: FieldValidator): InputValidationMixin {
        return RealInputValidationMixin(inputStream, validator)
    }
}
```

Othrewise we need to create a factory to create mixin, which ruins the purpose of the factory pattern

valentunn (approved):

PR #1182, opened 2023-10-25
Swap/max button

valentunn:
Basic idea:

* `MaxActionProvider` is  generic enough to handle any use-case with max available balance flow since its exposes just `Flow<Balance>`
* `MaxActionProvider` implementations are stackable. There are currently two implementations - one for just using field from `Asset`, another one deducts fee based on `GenericFeeLoaderMixin<F>`
* `BaseAmountChooserProvider` depends on `MaxActionProvider` and implements tracking for max available balances changes and updates its internal state with `InputKind.MAX_ACTION`
* Note that neither `BaseAmountChooserProvider` nor `MaxActionProvider` are responsible for updateing fee, they just update `inputState`. It is up for the screen (logic in viewModel) to define when exactly they want to recalculate fee. They might rely on `InputKind` to discinguish between regular and max button inputs

image

### DIFF
--- a/feature-wallet-api/src/main/java/io/novafoundation/nova/feature_wallet_api/presentation/mixin/amountChooser/BaseAmountChooserProvider.kt
+++ b/feature-wallet-api/src/main/java/io/novafoundation/nova/feature_wallet_api/presentation/mixin/amountChooser/BaseAmountChooserProvider.kt
@@ -162,7 +162,7 @@ open class BaseAmountChooserProvider(
 
         private fun MaxActionProvider?.maxAvailableForAction(): Flow<Balance?> = this?.maxAvailableForAction?.balance() ?: flowOf(null)
 
-        private fun MaxActionProvider?.maxAvailableForDisplay(): Flow<Balance?> = this?.maxAvailableForDisplay?: flowOf(null)
+        private fun MaxActionProvider?.maxAvailableForDisplay(): Flow<Balance?> = this?.maxAvailableForDisplay ?: flowOf(null)
 
         private fun Flow<MaxAvailableForAction?>.balance(): Flow<Balance?> = map { it?.balance }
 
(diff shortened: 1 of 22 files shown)


feature-wallet-api/src/main/java/io/novafoundation/nova/feature_wallet_api/presentation/mixin/amountChooser/BaseAmountChooserProvider.kt
```diff
@@ -1,28 +1,57 @@
...
 ) : AmountChooserMixinBase.Presentation,
     CoroutineScope by coroutineScope,
     WithCoroutineScopeExtensions by WithCoroutineScopeExtensions(coroutineScope) {
 
+    private val chainAssetFlow = tokenFlow.map { it?.configuration }
+        .distinctUntilChanged()
+
     final override val inputState = MutableStateFlow(defaultState())
 
+    @Deprecated("Use amountInput instead")
```
antonijzelinskij, line 54: Do you mean `Use amountState instead`?

feature-wallet-api/src/main/java/io/novafoundation/nova/feature_wallet_api/presentation/mixin/amountChooser/BaseAmountChooserProvider.kt
```diff
@@ -31,18 +60,115 @@ open class BaseAmountChooserProvider(
             inputState.map { input -> input.parseBigDecimalOrNull() }
         }.share()
 
-    override val amount: Flow<BigDecimal> = amountState
+    private val _amount: Flow<BigDecimal> = amountState
         .map { it.value.orZero() }
         .share()
 
+    @Deprecated("Use amountState instead")
+    override val amount: Flow<BigDecimal> = _amount
```
antonijzelinskij, line 68: Why do we use both variables if we can have only one?

antonijzelinskij (approved):

feature-wallet-api/src/main/java/io/novafoundation/nova/feature_wallet_api/presentation/mixin/amountChooser/BaseAmountChooserProvider.kt
valentunn: Yes

feature-wallet-api/src/main/java/io/novafoundation/nova/feature_wallet_api/presentation/mixin/amountChooser/BaseAmountChooserProvider.kt
valentunn: To avoid deprecation warnings inside BaseAmountChooserProvider itself

PR #1196, opened 2023-11-02
Changed rate change validatin + add validation for amount out

antonijzelinskij:
#86931508w
I also fixed the case where `amountIn` was being rounded to zero in `amountOut` by adding a positiveAmount validation for `amountOut` and changing the field validator.


stepanLav:
Task linked: [CU-86931508w Android: Change slippage validation](https://app.clickup.com/t/86931508w)

feature-swap-impl/src/main/java/io/novafoundation/nova/feature_swap_impl/domain/validation/validations/SwapRateChangesValidation.kt
```diff
@@ -1,38 +1,48 @@
...
 }
+
+private fun SwapLimit.calculateRateDifference(quotedBalance: Balance): Perbill {
+    val rateDifference = when (this) {
+        is SwapLimit.SpecifiedIn -> (amountOutMin.toBigDecimal() - quotedBalance.toBigDecimal()) / quotedBalance.toBigDecimal()
+        is SwapLimit.SpecifiedOut -> (quotedBalance.toBigDecimal() - amountInMax.toBigDecimal()) / quotedBalance.toBigDecimal()
+    }
+
+    return rateDifference.asPerbill()
+}
```
valentunn, line 48: The idea was a little bit different - we should not use division here but rather just compare `quotedBalance` with either `amountOutMin` or `amountInMax`

valentunn (changes requested):

valentunn (approved):

PR #1208, opened 2023-11-09
Using local runtime version

antonijzelinskij:


runtime/src/main/java/io/novafoundation/nova/runtime/extrinsic/ExtrinsicBuilderFactory.kt
```diff
@@ -55,7 +59,7 @@ class ExtrinsicBuilderFactory(
 
         val accountAddress = chain.addressOf(accountId)
 
-        val runtimeVersion = rpcCalls.getRuntimeVersion(chain.id)
+        val runtimeVersion = getRuntimeVersion(chain) ?: return emptySequence()
```
antonijzelinskij, line 62: Maybe we can call `rpcCalls.getRuntimeVersion(chain.id)` version here as a fallback instead of returning `emptySequence`

valentunn: Yeah thats a good idea. Lets, however, change request to be `state_getRuntimeVersion` instead of `chain_getRuntimeVersion`. We can create a separate request class similar to `RuntimeVersionRequest`

core-db/src/main/java/io/novafoundation/nova/core_db/migrations/50_51_AddTransactionVersionToRuntime.kt
```diff
@@ -0,0 +1,11 @@
+package io.novafoundation.nova.core_db.migrations
+
+import androidx.room.migration.Migration
+import androidx.sqlite.db.SupportSQLiteDatabase
+
+val AddTransactionVersionToRuntime_50_51 = object : Migration(50, 51) {
+
+    override fun migrate(database: SupportSQLiteDatabase) {
+        database.execSQL("ALTER TABLE chain_runtimes ADD COLUMN transactionVersion INTEGER")
```
valentunn, line 9: Have you tested this migration? I have concerns it wont work without explicitly specifying `DEFAULT NULL`

valentunn (approved):

core-db/src/main/java/io/novafoundation/nova/core_db/migrations/50_51_AddTransactionVersionToRuntime.kt
antonijzelinskij: Yes the migration is successful but I checked why: It seems that SQL use default value here and for integer it's 0, So it's good you noticed this issue. I'll fix it

valentunn (approved):

PR #1226, opened 2023-11-14
Fix/transactions sync fail

antonijzelinskij:
#86934c9eb


core-db/src/main/java/io/novafoundation/nova/core_db/dao/OperationDao.kt
```diff
@@ -163,6 +163,7 @@ abstract class OperationDao {
             is PoolRewardTypeLocal -> insertPoolRewardType(type)
             is SwapTypeLocal -> insertSwapType(type)
             is TransferTypeLocal -> insertTransferType(type)
+            else -> {}
```
antonijzelinskij, line 166: Idk why but now compiler requires else here :D

valentunn (approved):

PR #1228, opened 2023-11-15
Fix - false positive VoteType si mapping for Interlay

valentunn:
Interlay uses forked version of  `democracy` pallet that has custom `Vote` type that doesnt have convitction: https://github.com/interlay/interbtc/blob/94066c995907506454522e49b02180bc35692b52/crates/democracy/src/types.rs#L15
Current implement of VoteSiTypeMapping triggers on this type due to its path (not empty + last segment is "Vote")
While we dont support Interlay governance, users can use it from Dapp browser. In such case this issue prevents correct decoding of extrinsic and thus fee loading fails
Fix uses full path matching for known Vote types

#86932gf41


stepanLav:
Task linked: [CU-86932gf41 Android. Can not calculate fee for sign governance vote tx](https://app.clickup.com/t/86932gf41)

antonijzelinskij (approved):

PR #1234, opened 2023-11-15
Fix/considering self sufficient validation

antonijzelinskij:
#869347jdh
- Fixed swap localization for operations
- Fixed validation which should consider minimal balance if we send tokens to non sufficient asset
- Fixed validation text that for feeAsset.transferable > fee + ed
- Fixed validation text that for feeAsset.transferable > fee 
Old text: You can use up to %s since you need to pay\n%s for network fee.
New one: Sorry, you don\'t have enough funds to pay the network fee.


stepanLav:
Task linked: [CU-869347jdh Android. Wrong validation text](https://app.clickup.com/t/869347jdh)

feature-swap-impl/src/main/java/io/novafoundation/nova/feature_swap_impl/domain/interactor/SwapInteractor.kt
```diff
@@ -139,6 +139,8 @@ class SwapInteractor(
 
             positiveAmountOut()
 
+            sufficientBalanceInFeeAsset()
```
antonijzelinskij, line 142: changed validation index since it should stay before checking feeAsset.transferable > fee + ed

valentunn (approved):

antonijzelinskij:
Fixed sufficient balance to pay fee validation in case when nativeAsset = feeAsset

valentunn (approved):

PR #1263, opened 2023-12-05
Feature/light sync

valentunn:


ERussel (commented):
Looks good! As future improvements it worth to consider the following:

0) When chain goes to full sync mode it doesn't switch back to light sync anymore. It would be helpful to support this for cases when user's balance goes to zero or wallet is removed.

1) Chain connection state is currently depends on current wallet. Instead we could make batch subscription and listen system account updates for all wallets. In that case we will not have an overhead when selected wallet changes.

ERussel (approved):

antonijzelinskij (approved):

core-db/src/main/java/io/novafoundation/nova/core_db/migrations/52_53_AddConnectionStateToChains.kt
```diff
@@ -0,0 +1,25 @@
...
+val AddConnectionStateToChains_53_54 = object : Migration(53, 54) {
+    override fun migrate(database: SupportSQLiteDatabase) {
+        val defaultConnectionState = ChainLocal.Default.CONNECTION_STATE_DEFAULT
+
+        database.execSQL("ALTER TABLE chains ADD COLUMN connectionState TEXT NOT NULL DEFAULT '$defaultConnectionState'")
+
+        // Enable full for chains that have some balance synced
+        database.execSQL(
+            """
+            UPDATE chains SET connectionState = 'FULL_SYNC'
```
ERussel, line 16: From this logic it mean that noRuntime (evm) netowrks might work as light_sync or full_sync. Probably it is better to manually set light_sync or full_sync for them.

valentunn:
> Looks good! As future improvements it worth to consider the following:
> 
> 0. When chain goes to full sync mode it doesn't switch back to light sync anymore. It would be helpful to support this for cases when user's balance goes to zero or wallet is removed.
> 1. Chain connection state is currently depends on current wallet. Instead we could make batch subscription and listen system account updates for all wallets. In that case we will not have an overhead when selected wallet changes.

On point 1 - thought about it as well but decided to not implement it since I thought that once user used certain chain (whether by getting a balance or accessing its runtime in some way) he/she might be interested in it even if account is wiped. So its better to discuss it separately
On poin 2 - imho it only produces an overhead if Point 1 is implemented. I also see it might potentially increase the load in case user have many accounts

PR #1267, opened 2023-12-13
Fix/staking paged exposures

valentunn:
* Support PagedExposures for main staking logic
* Make MaxRewardedNominatorsPerValidator optional
* Support PagedExposures for payouts, including support for migration period
* Refactor `PayoutRepository` to use `StorageDataSource`
* Better type-safety for `multi` queries
* Do not sync historical EraRewards and EraPoints (HistoricalUpdateMediator)


feature-staking-impl/src/main/java/io/novafoundation/nova/feature_staking_impl/domain/StakingInteractor.kt
```diff
@@ -327,13 +328,14 @@ class StakingInteractor(
     private suspend fun activeNominators(chainId: ChainId, exposures: Collection<Exposure>): Int {
         val activeNominatorsPerValidator = stakingConstantsRepository.maxRewardedNominatorPerValidator(chainId)
 
-        return exposures.fold(mutableSetOf<AccountIdKey>()) { acc, exposure ->
-            acc += exposure.others.sortedByDescending(IndividualExposure::value)
-                .take(activeNominatorsPerValidator)
-                .map { it.who.intoKey() }
-
-            acc
-        }.size
```
valentunn, line 336: Noticed quite an unefficient computuation of active nominators count

feature-wallet-api/src/main/java/io/novafoundation/nova/feature_wallet_api/presentation/mixin/amountChooser/AmountChooserMixin.kt
```diff
@@ -42,7 +42,6 @@ interface AmountChooserMixinBase : CoroutineScope {
 
     interface Presentation : AmountChooserMixinBase {
 
-        @Deprecated("Use amountState instead")
```
valentunn, line 45: Removed deprecation since multiple screens will use what `amount` offers anyway

runtime/src/main/java/io/novafoundation/nova/runtime/storage/source/query/StorageQueryContext.kt
```diff
@@ -103,9 +107,19 @@ interface StorageQueryContext {
...
 
-fun Map<StorageEntry, Map<StorageKeyComponents, Any?>>.singleValueOf(storageEntry: StorageEntry) = getValue(storageEntry).values.first()
+@Suppress("DEPRECATION")
+@OptIn(ExperimentalContracts::class)
+suspend fun StorageQueryContext.multi(
+    builderBlock: MultiQueryBuilder.() -> Unit
+): MultiQueryBuilder.Result {
+    contract {
+        callsInPlace(builderBlock, InvocationKind.EXACTLY_ONCE)
+    }
```
valentunn, line 117: We cannot define a contract in the interface so this is a workaround to help compiler understand we're invoking `builderBlock` at least once so compiler will understand that assignments present inside `builderBlock` are actually hapenning  to avoid `variable must be itialized` compilation error

antonijzelinskij (approved):

PR #1268, opened 2023-12-13
Fix evm tx history crash

valentunn:
Fix crash when there is no function saved for the contract call

#8693ag0ax

image


stepanLav:
Task linked: [CU-8693ag0ax Android. Crash for SAMA tx history](https://app.clickup.com/t/8693ag0ax)

antonijzelinskij (approved):

PR #1286, opened 2023-12-21
Feature/proxy fee alert

antonijzelinskij:
#8693b2kqe


stepanLav:
Task linked: [CU-8693b2kqe Android. Proxy doesn't have enough fee alert](https://app.clickup.com/t/8693b2kqe)

app/src/main/java/io/novafoundation/nova/app/root/presentation/requestBusHandler/ProxyExtrinsicValidationRequestBusHandler.kt
```diff
@@ -0,0 +1,64 @@
...
+
+    private fun ValidationSystemBuilder<ProxiedExtrinsicValidationPayload, ProxiedExtrinsicValidationFailure>.proxyHasEnoughFee() {
+        proxyHasEnoughFeeValidation(
+            factory = proxyHaveEnoughFeeValidationFactory,
+            proxyAccountId = { it.proxyAccountId },
+            call = { it.call },
+            chainWithAsset = { it.chainWithAsset },
+            proxyNotEnoughFee = { payload, availableBalance, fee ->
+                val errorMessage = buildNotEnoughFeeMessage(payload, availableBalance, fee)
+                ProxiedExtrinsicValidationFailure.ProxyNotEnoughFee(errorMessage)
```
antonijzelinskij, line 50: I didn't find that we send error messages via failure, but it seems this is the simplest way to implement alert, since in account module we don't have amountFromPlanks and formatTokenAmount funs

core-db/src/main/java/io/novafoundation/nova/core_db/dao/MetaAccountDao.kt
```diff
@@ -188,6 +188,23 @@ interface MetaAccountDao {
...
+            UNION ALL
+            SELECT pa.proxyMetaId
+            FROM proxy_accounts pa
+            INNER JOIN Sequence s ON s.proxyMetaId = pa.proxiedMetaId
+        )
+        SELECT MAX(proxyMetaId) as finalMetaId
+        FROM Sequence
+    """
+    )
+    fun getLastProxyAccountId(proxiedMetaId: Long): Long?
```
antonijzelinskij, line 207: This recursion find root proxy account in all chain. It works nice for non-nested proxy-accounts, but I didn't check how it works for nested, I guess it would be also good

feature-account-impl/src/main/java/io/novafoundation/nova/feature_account_impl/data/signer/proxy/ProxiedSigner.kt
```diff
@@ -60,14 +70,20 @@ class ProxiedSigner(
...
 
     override suspend fun signExtrinsic(payloadExtrinsic: SignerPayloadExtrinsic): SignedExtrinsic {
+        val chain = chainRegistry.getChain(payloadExtrinsic.chainId)
         val proxyMetaAccount = getProxyMetaAccount()
 
         acknowledgeProxyOperation(proxyMetaAccount)
 
+        if (isRootProxiedSigner()) {
+            validateExtrinsic(payloadExtrinsic, chain)
+        }
```
antonijzelinskij, line 80: Checking that signer is root, to avoid validation in nested proxied signers

feature-account-impl/src/main/java/io/novafoundation/nova/feature_account_impl/data/signer/proxy/ProxiedSigner.kt
```diff
@@ -78,9 +94,32 @@ class ProxiedSigner(
...
+        val proxyAccountId = proxyAccount.accountIdIn(chain) ?: throw IllegalStateException("Proxy account is not found")
+
+        val callInstance = signedExtrinsic.call.toCallInstance() ?: signingNotSupported()
+
+        val validationPayload = ProxiedExtrinsicValidationPayload(
+            proxyAccount,
+            proxyAccountId,
+            ChainWithAsset(chain, chain.commissionAsset),
+            callInstance.call
+        )
```
antonijzelinskij, line 108: Unfortunately we should to pass call instance instead of calculating fee here, because we can't inject ExtrinsicService here due to cyclical dependencies

valentunn (approved):

PR #1294, opened 2023-12-25
Proxy/fee validation

valentunn:
#869379g01

Basic idea: 
1. Extend `Fee` interface to hold information who will actually sign the transaction by adding field of type `SubmissionOrigin`. This requries extending `FeeSigner` interface to have ` actualFeeSignerId(chain: Chain)` and ` requestedFeeSignerId(chain: Chain` since `signerAccountId(chain: Chain)` returns fake id generated by fake keypair that will be used for signing
2. Migrate all screens to use not-so-ago added `DecimalFee` instead of `BigDecimal`
3. Modify validations that use fee to consider `submissionOrigin` by using `amountByRequestedAccount` instead of `amount`
4. Handle a few specific cases separately like Swaps, External Sign (Dapp browser + WC) and transaction history insertion

I've also replace `requireFee` with `awaitFee` with former being the outdated behavior


stepanLav:
Task linked: [CU-869379g01 Android. Support proxy in fee validations](https://app.clickup.com/t/869379g01)

PR #1303, opened 2023-12-28
Per chain proxy sync

valentunn:
Removed timeout since 10 seconds seems to be not enough right after app start


ERussel (approved):

feature-account-impl/src/main/java/io/novafoundation/nova/feature_account_impl/data/proxy/RealProxySyncService.kt
```diff
@@ -21,21 +21,19 @@ import io.novafoundation.nova.feature_account_api.domain.model.MetaAccount
...
-import io.novafoundation.nova.feature_account_impl.data.repository.addAccount.proxied.ProxiedAddAccountRepository.Payload
 import io.novafoundation.nova.runtime.multiNetwork.ChainRegistry
 import io.novafoundation.nova.runtime.multiNetwork.chain.model.Chain
 import io.novafoundation.nova.runtime.multiNetwork.chain.model.ChainId
 import io.novafoundation.nova.runtime.multiNetwork.findChains
 import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.launch
-import kotlinx.coroutines.withTimeoutOrNull
 
 private const val SYNC_TIMEOUT = 10_000L // 10 seconds
```
antonijzelinskij, line 31: This should be removed

PR #1304, opened 2023-12-29
Fix - crash for proxied receive

valentunn:
#8693djk9c
We dont need to use public key here
image


stepanLav:
Task linked: [CU-8693djk9c Android. Fix QR format for proxied accounts](https://app.clickup.com/t/8693djk9c)

ERussel (approved):

PR #1311, opened 2024-01-04
Selecting first meta account in case when selected proxied is removed due to deactivation

antonijzelinskij:
#8693erxyh


stepanLav:
Task linked: [CU-8693erxyh Android. Process case when remove proxy from selected proxied wallet](https://app.clickup.com/t/8693erxyh)

ERussel (approved):

PR #1313, opened 2024-01-09
Fix/proxy ledger

valentunn:
#8693fp8fu

Ledger flow used selected meta account to lookup derivation path, which will point to proxied that doesn't have ledger derivation path if proxied wallet is used. 
I've propagated actual meta account to the shared signing state for separate flow signing
I've also fixed signature verification for Polkadot Vault which had the same issue


stepanLav:
Task linked: [CU-8693fp8fu Android. Ledger operation failed on proxy calls](https://app.clickup.com/t/8693fp8fu)

antonijzelinskij (approved):

PR #1317, opened 2024-01-10
Fix - robonomics demoracy unlock call encoding

valentunn:
#8693g6ehg

Use type from the call metadata


ERussel (approved):

stepanLav:
Task linked: [CU-8693g6ehg Android. Robonomics Unlock fee calculation fails](https://app.clickup.com/t/8693g6ehg)

PR #1318, opened 2024-01-10
Revisit usages of all accounts

valentunn:
The only place where we still want to use all accounts is `updatedProxieds`
#8693fzyqz


stepanLav:
Task linked: [CU-8693fzyqz Android. Check for deactivated wallets corner cases everywhere](https://app.clickup.com/t/8693fzyqz)

PR #1320, opened 2024-01-10
Fix/staking reactive updates for relaychain staking

valentunn:
#8693g0ewg
#8693g0jn4

We use `eraCalculatorFlow`, which is endless flow, as an inner flow inside `combineTransform` for unbondings and stake summary when staking state is "waiting for next era". Howerver, `combineTransform` does not process next upstream item until inner flow completes, so in case of "waiting for next era" no updates after initial state were processed

Fixed by replacing `combineTransform` with `combineToPair(...).flatMapLatest{...}`


stepanLav:
Task linked: [CU-8693g0ewg Android. Inactive state after start staking on Westend](https://app.clickup.com/t/8693g0ewg)

antonijzelinskij (approved):

PR #1339, opened 2024-01-19
Nested proxies

antonijzelinskij:
#8693br4aj
#8693p5ngf
OnChain state:
image

Accounts in Nova Wallet:

After remove "Staking" proxy:


stepanLav:
Task linked: [CU-8693br4aj Android. Support nested proxies](https://app.clickup.com/t/8693br4aj)

feature-account-impl/src/main/java/io/novafoundation/nova/feature_account_impl/data/signer/proxy/ProxiedSigner.kt
```diff
@@ -73,21 +73,30 @@ class ProxiedSigner(
...
-            validateExtrinsic(payloadExtrinsic, chain)
+            validateExtrinsic(payloadToSign, chain)
         }
 
         val delegate = createDelegate(proxyMetaAccount)
-        val modifiedPayload = modifyPayload(proxyMetaAccount, payloadExtrinsic, chain)
-
-        val signedExtrinsic = delegate.signExtrinsic(modifiedPayload)
-        return signedExtrinsic
+        return delegate.signExtrinsic(payloadToSign)
```
valentunn, line 99: I dont quite like the idea of letting `signExtrinsic` handle both final and non-final payloads. It might quickly become error pronse as we add more signer types like multisig. `modifyPayload` can return both "one layer deep" wrapped payload and "final payload" so `signExtrinsic` might use final payload for validation and "one layer deep" to pass it to the bottom layer.

feature-proxy-api/src/main/java/io/novafoundation/nova/feature_proxy_api/domain/model/ProxyType.kt
```diff
@@ -1,24 +1,24 @@
 package io.novafoundation.nova.feature_proxy_api.domain.model
 
-sealed class ProxyType(val name: String) {
+sealed class ProxyType(val name: String, val controllableFrom: List<ProxyType>) {
```
valentunn, line 3: Ideally this shouldnt be embedded into ProxyType itself since it is actually what `ProxtCallFilter` is responsible for. Refactoring it will require some efforts though so we can leave it as it is rn

runtime/src/main/java/io/novafoundation/nova/runtime/extrinsic/signer/NovaSigner.kt
```diff
@@ -23,4 +26,8 @@ interface FeeSigner : NovaSigner {
      * It might not be equal to [actualFeeSignerId] if [Signer] modifies the payload
      */
     suspend fun requestedFeeSignerId(chain: Chain): AccountId
+
+    override suspend fun modifyPayload(payloadExtrinsic: SignerPayloadExtrinsic): SignerPayloadExtrinsic {
+        throw NotImplementedError("This method should not be called")
+    }
```
valentunn, line 32: Looks like it is reasonable now to create a separate interface for real signers to expose this method there

valentunn (approved):
Looks good, comments are non-blocking for merge

antonijzelinskij:
Fixed creating proxied accounts for this state:
image

feature-account-impl/src/main/java/io/novafoundation/nova/feature_account_impl/data/signer/proxy/ProxiedSigner.kt
```diff
@@ -89,14 +89,12 @@ class ProxiedSigner(
...
         }
 
-        val payloadToSign = if (isRootProxied) modifyPayload(payloadExtrinsic) else payloadExtrinsic
-
         // TODO this wont use the actual payload for fee validation when multiple nested proxies are used
         // We need to design a universal solution
         // We actually can use `signedExtrinsic.payload` to access actual payload but in this case validation will happen only after signing
         // which will have bad UX with Vault and Ledger.
         // As an option we could separate signing and wrapping step specifically for such nested signers and use only the wrapping step before fee validation
-        val modifiedPayload = modifyPayload(proxyMetaAccount, payloadExtrinsic, chain)
```
valentunn: Should we remove this comment?

valentunn (approved):

feature-account-impl/src/main/java/io/novafoundation/nova/feature_account_impl/data/signer/proxy/ProxiedSigner.kt
antonijzelinskij: Yes, I did it

PR #1340, opened 2024-01-19
Prepare identity to runtime upgrade

antonijzelinskij:
#8693k308m


feature-account-impl/src/main/java/io/novafoundation/nova/feature_account_impl/data/network/blockchain/bindings/Identity.kt
```diff
@@ -15,20 +15,25 @@ import jp.co.soramitsu.fearless_utils.extensions.toHexString
...
 
 @UseCaseBinding
-fun bindIdentity(dynamic: Any?,): OnChainIdentity? {
+fun bindIdentity(dynamic: Any?): OnChainIdentity? {
     if (dynamic == null) return null
 
     val decoded = dynamic.castToStruct()
 
-    val identityInfo = decoded.get<Struct.Instance>("info") ?: incompatible()
+    val identityInfo = decoded.get<Struct.Instance>("registration")
```
valentunn, line 23: As far as I understood it is not a struct but a tuple (registration, userName), thus it will be a list

valentunn (approved):

PR #1341, opened 2024-01-19
7.7.2 to develop

antonijzelinskij:


runtime/src/main/java/io/novafoundation/nova/runtime/multiNetwork/runtime/RuntimeSyncService.kt
```diff
@@ -97,12 +97,13 @@ class RuntimeSyncService(
...
-                val syncResult = sync(chainId, forceFullSync)
-                syncResult?.let { _syncStatusFlow.emit(it) }
-            }
+            val syncResult = runCatching {
+                sync(chainId, forceFullSync)
+            }.getOrNull()
 
             syncFinished(chainId)
+
+            syncResult?.let { _syncStatusFlow.emit(it) }
```
antonijzelinskij, line 106: It's already on dev. It seems that it is showing since I use cherry pick

valentunn (approved):

PR #1346, opened 2024-01-23
Support spend_local call

antonijzelinskij:
#8693m3fh8


stepanLav:
Task linked: [CU-8693m3fh8 Android. Support new treasury storage](https://app.clickup.com/t/8693m3fh8)

feature-governance-impl/src/main/java/io/novafoundation/nova/feature_governance_impl/domain/referendum/details/call/treasury/TreasurySpendAdapter.kt
```diff
@@ -15,7 +15,8 @@ class TreasurySpendAdapter : ReferendumCallAdapter {
         call: GenericCall.Instance,
         context: ReferendumCallParseContext
     ): ReferendumCall? {
-        if (!call.instanceOf(Modules.TREASURY, "spend")) return null
+        // TODO: spend call is using MultiLocation now instead of MultiAddress so binding throws an exception
```
ERussel, line 18: Is this a problem that spend call now throws an exception? We want still display such referendums but without amount?

ERussel (approved):

PR #1352, opened 2024-01-26
Hydradx/referrals

valentunn:
Track user's current referral and add Nova referral registration to swap call is user doesnt have one
Support batch calls in realtime tx history

### DIFF
--- a/runtime/src/test/java/io/novafoundation/nova/runtime/extrinsic/visitor/impl/BatchAllWalkTest.kt
+++ b/runtime/src/test/java/io/novafoundation/nova/runtime/extrinsic/visitor/impl/BatchAllWalkTest.kt
@@ -157,11 +157,11 @@ internal class BatchAllWalkTest {
             addAll(innerBatchEvents)
             add(itemCompleted())
 
-                // second level batch starts
+            run {
                 addAll(innerBatchEvents)
                 add(itemCompleted())
 
-                    // third level batch starts
+                run {
                     addAll(innerBatchEvents)
                     add(itemCompleted())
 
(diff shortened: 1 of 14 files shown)


runtime/src/test/java/io/novafoundation/nova/runtime/extrinsic/visitor/impl/BatchAllWalkTest.kt
```diff
@@ -0,0 +1,234 @@
...
+                    // third level batch ends
+
+                add(itemCompleted())
+                // second level batch ends
+                add(batchCompleted())
+
+            add(itemCompleted())
+
+            addAll(innerBatchEvents)
+            add(itemCompleted())
```
antonijzelinskij, line 181: I suggest to use `kotlin.run {  }` to highlight levels instead of extra spaces

runtime/src/test/java/io/novafoundation/nova/runtime/extrinsic/visitor/impl/Common.kt
```diff
@@ -0,0 +1,132 @@
...
+    val testEvent = mockEvent(testModule.name, "test")
+
+    operator fun component1(): GenericCall.Instance {
+        return testInnerCall
+    }
+
+    operator fun component2():GenericEvent.Instance {
+        return testEvent
+    }
+}
```
antonijzelinskij, line 118: Why did we describe this class and `createTestModuleWithCall` here?
It seems better to move it to the test directory

antonijzelinskij (approved):

runtime/src/test/java/io/novafoundation/nova/runtime/extrinsic/visitor/impl/Common.kt
valentunn: It is in the test directory already

PR #1380, opened 2024-02-14
Hydra dx swaps base

valentunn:


feature-swap-impl/src/main/java/io/novafoundation/nova/feature_swap_impl/data/assetExchange/hydraDx/omnipool/model/DynamicFee.kt
antonijzelinskij, line 1: DynamicFee is placed in the hydraDx directory. Would it be better to call it HydraSwapFee?

antonijzelinskij (approved):

feature-swap-impl/src/main/java/io/novafoundation/nova/feature_swap_impl/data/assetExchange/hydraDx/omnipool/model/DynamicFee.kt
valentunn: I've used the same name as in blockhain. I think its ok given that it is under hydra package so it is related to hydra only

PR #1381, opened 2024-02-15
Wallet connect deeplinks

valentunn:


app/src/main/java/io/novafoundation/nova/app/root/presentation/deepLinks/RootDeepLinkHandler.kt
```diff
@@ -17,7 +17,11 @@ class RootDeepLinkHandler(
     }
 
     override suspend fun handleDeepLink(data: Uri) {
-        nestedHandlers.find { it.matches(data) }
-            ?.handleDeepLink(data)
+        runCatching {
```
antonijzelinskij, line 20: Maybe it would be nice to show error message if we catch an exception

antonijzelinskij (approved):

PR #1382, opened 2024-02-15
Do not automatically switch to asset in if it cant be accepted as a payment asset

valentunn:
#8693um1kc


stepanLav:
Task linked: [CU-8693um1kc Android. Exclude pool tokens from setting to fee currencies](https://app.clickup.com/t/8693um1kc)

antonijzelinskij (approved):

PR #1387, opened 2024-02-16
Implemented select multiple wallets fragment

antonijzelinskij:
#8693rppwu

### DIFF
--- a/common/src/main/java/io/novafoundation/nova/common/view/recyclerview/adapter/text/TextAdapter.kt
+++ b/common/src/main/java/io/novafoundation/nova/common/view/recyclerview/adapter/text/TextAdapter.kt
@@ -1,0 +1,54 @@
+package io.novafoundation.nova.common.view.recyclerview.adapter.text
+
+import android.view.View
+import android.view.ViewGroup
+import androidx.recyclerview.widget.ListAdapter
+import androidx.annotation.StyleRes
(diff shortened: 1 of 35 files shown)


stepanLav:
Task linked: [CU-8693rppwu Android. Notifications. Select wallet settings](https://app.clickup.com/t/8693rppwu)

common/src/main/java/io/novafoundation/nova/common/view/recyclerview/adapter/text/TextAdapter.kt
```diff
@@ -0,0 +1,54 @@
...
+import androidx.recyclerview.widget.DiffUtil
+import androidx.recyclerview.widget.RecyclerView.ViewHolder
+import io.novafoundation.nova.common.R
+import io.novafoundation.nova.common.utils.inflateChild
+import kotlinx.android.synthetic.main.item_text.view.itemText
+
+class TextAdapter(
+    initText: String? = null,
+    @StyleRes private val styleRes: Int = R.style.TextAppearance_NovaFoundation_Bold_Title2
+) : ListAdapter<TextRvItem, TextViewHolder>(TextAdapterDiffCallback()) {
```
valentunn, line 16: What is the reason to use ListAdapter here for 1 element? Can we use something similar like we do in `DAppHeaderAdapter`?

valentunn (approved):

PR #1392, opened 2024-02-19
Fix/push handling

antonijzelinskij:


feature-push-notifications/src/main/java/io/novafoundation/nova/feature_push_notifications/data/NovaFirebaseMessagingService.kt
```diff
@@ -1,22 +1,35 @@
...
+import kotlinx.coroutines.SupervisorJob
+import kotlinx.coroutines.cancel
+import kotlinx.coroutines.launch
 import kotlinx.coroutines.tasks.await
+import kotlin.coroutines.CoroutineContext
 
-class NovaFirebaseMessagingService : FirebaseMessagingService() {
+class NovaFirebaseMessagingService : FirebaseMessagingService(), CoroutineScope {
+
+    override val coroutineContext: CoroutineContext = Dispatchers.Main + SupervisorJob()
```
valentunn, line 22: Should we really run on the main dispatcher instead of Default or IO?

feature-push-notifications/src/main/java/io/novafoundation/nova/feature_push_notifications/data/data/settings/model/VersionedPushSettingsCache.kt
```diff
@@ -0,0 +1,6 @@
+package io.novafoundation.nova.feature_push_notifications.data.data.settings.model
+
+class VersionedPushSettingsCache(
+    val version: String,
+    val settings: String // json here
```
valentunn, line 5: Instead of comment we can create a typealias:

``` typelias Json = String ```

File: feature-push-notifications/src/main/java/io/novafoundation/nova/feature_push_notifications/data/data/subscription/RealPushSubscriptionService.kt [hunk 3]
```diff
@@ -167,4 +170,19 @@ class RealPushSubscriptionService(
...
+
+    private fun List<ChainId>.transfromChainIdsTo16Hex(): List<String> {
+        return map { chainId -> chainId.to16Hex() }
+    }
+
+    private fun ChainId.to16Hex(): String {
+        return fromHex()
+            .take(16)
+            .toByteArray()
+            .toHexString(withPrefix = true)
```
valentunn, line 186: We can reduce number of operations by doing `removeHexPrefix.().take(32).requireHexPrefix()`. Also iirc similar code but without final prefix is used in `Caip2Identifier.Polkadot` so we can create a common function to reuse it in both places

feature-push-notifications/src/main/java/io/novafoundation/nova/feature_push_notifications/data/data/subscription/RealPushSubscriptionService.kt
```diff
@@ -167,4 +170,19 @@ class RealPushSubscriptionService(
...
+
+    private fun <T> Map<ChainId, T>.transfromChainIdsTo16Hex(): Map<String, T> {
+        return mapKeys { (chainId, _) -> chainId.to16Hex() }
+    }
+
+    private fun List<ChainId>.transfromChainIdsTo16Hex(): List<String> {
+        return map { chainId -> chainId.to16Hex() }
+    }
+
+    private fun ChainId.to16Hex(): String {
```
valentunn, line 182: I suggest to use `hexPrefix16()`. When first reading `to16Hex` is some kind of base but not the prefix length

feature-push-notifications/src/main/java/io/novafoundation/nova/feature_push_notifications/data/presentation/handling/BaseNotificationHandler.kt
```diff
@@ -0,0 +1,35 @@
...
+                resourceManager.getString(channelNameRes),
+                importance
+            )
+            notificationManager.createNotificationChannel(channel)
+        }
+
+        return handleNotificationInternal(notificationManager, channelId, message)
+    }
+
+    protected abstract fun handleNotificationInternal(notificationManager: NotificationManagerCompat, channelId: String, message: RemoteMessage): Boolean
```
valentunn, line 34: Imho we can make `notificationManager` protected in the class and access it that way in subclasses instead

feature-push-notifications/src/main/java/io/novafoundation/nova/feature_push_notifications/data/presentation/handling/SendingNotificationHandler.kt
```diff
@@ -0,0 +1,35 @@
...
+
+class SendingNotificationHandler(
+    private val context: Context,
+    notificationManager: NotificationManagerCompat,
+    resourceManager: ResourceManager,
+) : BaseNotificationHandler(
+    notificationManager,
+    resourceManager,
+    R.string.default_notification_channel_id,
+    R.string.default_notification_channel_name,
```
valentunn, line 18: Do we need to pass them each time? If we want it to be customizable and avoid passing default value many times we can use default parameters in `BaseNotificationHandler`

feature-push-notifications/src/main/java/io/novafoundation/nova/feature_push_notifications/data/presentation/handling/SendingNotificationHandler.kt
```diff
@@ -0,0 +1,35 @@
...
+) {
+
+    override fun handleNotificationInternal(notificationManager: NotificationManagerCompat, channelId: String, message: RemoteMessage): Boolean {
+        // TODO check if the notification has a send type
+
+        val notification = NotificationCompat.Builder(context, channelId)
+            .setContentTitle("\uD83D\uDCB8 Sent [Account name/id]") // TODO stub
+            .setContentText("Sent 0.5 ETH (\$900) to 0xc6c...81cdd on Moonbeam") // TODO stub
```
valentunn, line 26: Pending TODOs. I suggest to mark future TODO with the task number if possible so during review we can check if it is a forgotten todo or the inentional one

feature-push-notifications/src/main/java/io/novafoundation/nova/feature_push_notifications/data/presentation/handling/SendingNotificationHandler.kt
```diff
@@ -0,0 +1,35 @@
...
+            .setSmallIcon(R.drawable.ic_nova)
+            .setPriority(NotificationCompat.PRIORITY_HIGH)
+            .build()
+
+        notificationManager.notify(1, notification)
```
valentunn, line 31: `1` looks like a magic number here

valentunn (approved):

feature-push-notifications/src/main/java/io/novafoundation/nova/feature_push_notifications/data/presentation/handling/SendingNotificationHandler.kt
antonijzelinskij: Yes, I would like to use multiple notification channels depending on the push type

feature-push-notifications/src/main/java/io/novafoundation/nova/feature_push_notifications/data/presentation/handling/SendingNotificationHandler.kt
antonijzelinskij: Okay, I'll do this in the next time!
But in this case I'm sure we will not forget about this since we still need to handle other push types and I'll return to this task anyway

PR #1417, opened 2024-03-11
Fix/push technical debts

antonijzelinskij:
- Show push settings only if google services enabled
- Handle case when we disable push notifications in settings
- Fixed save button state
- Create intents for push notifications to open screens
- Add channels
- Support push-available for chains
- Fixed chooser intent
- Add MUATBLE flag for push notification
- Sync push settings after change account and remove meta account
- Select wallet before push open
- Fixed subscription handling to firebase topics
- Clean code

#86940qh22
#86941vgpe


feature-push-notifications/src/main/java/io/novafoundation/nova/feature_push_notifications/data/domain/interactor/PushNotificationsInteractor.kt
```diff
@@ -50,4 +56,31 @@ class RealPushNotificationsInteractor(
...
+
+    override suspend fun onMetaAccountRemoved(metaId: Long) {
+        val notificationsEnabled = pushSettingsProvider.isPushNotificationsEnabled()
+        val noAccounts = accountRepository.getActiveMetaAccountsQuantity() == 0
+        val pushSettings = pushSettingsProvider.getPushSettings()
+
+        if (notificationsEnabled && noAccounts) {
+            pushNotificationsService.updatePushSettings(false, null)
+        } else if (noAccounts) {
+            pushSettingsProvider.updateSettings(null)
```
valentunn, line 80: Why do we update notification settings to null while they should already be disabled? I think that's one of the places where logic error would become visible if we use sealed class for settings internal state

feature-push-notifications/src/main/java/io/novafoundation/nova/feature_push_notifications/data/presentation/handling/NotificationHandlerExt.kt
```diff
@@ -73,28 +73,37 @@ fun Context.makePendingIntent(intent: Intent): PendingIntent {
         this,
         PEDDING_INTENT_REQUEST_CODE,
         intent,
-        PendingIntent.FLAG_UPDATE_CURRENT
+        PendingIntent.FLAG_UPDATE_CURRENT or PendingIntent.FLAG_IMMUTABLE
```
valentunn, line 76: Can you please remind me how FLAG_IMMUTABLE affects the behavior of pending intent for our case?

valentunn (changes requested):

stepanLav:
Task linked: [CU-86940qh22 Android. tokenReceived and tokenSend is not updating](https://app.clickup.com/t/86940qh22)

feature-account-impl/src/main/java/io/novafoundation/nova/feature_account_impl/data/repository/addAccount/BaseAddAccountRepository.kt
```diff
@@ -2,16 +2,32 @@ package io.novafoundation.nova.feature_account_impl.data.repository.addAccount
...
 
-    final override suspend fun addAccount(payload: T): Long {
-        val metaId = addAccountInternal(payload)
+    final override suspend fun addAccount(payload: T): AddAccountResult {
+        val addAccountResult = addAccountInternal(payload)
         proxySyncService.startSyncing()
-        return metaId
+
+        // We should sync push notification settings after adding/changing a chain account
+        // to avoid notifications about status changes of unused accounts
```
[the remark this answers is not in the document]
antonijzelinskij: Agree. Fixed3

feature-account-impl/src/main/java/io/novafoundation/nova/feature_account_impl/data/repository/addAccount/secrets/SecretsAddAccountRepository.kt
```diff
@@ -17,16 +19,11 @@ abstract class SecretsAddAccountRepository<T>(
...
-) : BaseAddAccountRepository<T>(proxySyncService) {
+    proxySyncService: ProxySyncService,
+    metaAccountChangesEventBus: MetaAccountChangesEventBus
+) : BaseAddAccountRepository<T>(proxySyncService, metaAccountChangesEventBus) {
 
-    class Payload(
-        val derivationPaths: AdvancedEncryption.DerivationPaths,
-        val addAccountType: AddAccountType,
-        val accountSource: AccountSecretsFactory.AccountSource
-    )
```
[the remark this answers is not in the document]
antonijzelinskij: Yes, child classes has their own payloads

feature-push-notifications/src/main/java/io/novafoundation/nova/feature_push_notifications/data/data/PushNotificationsService.kt
```diff
@@ -23,14 +29,18 @@ interface PushNotificationsService {
 
     suspend fun initPushNotifications(): Result<Unit>
 
-    suspend fun updatePushSettings(enabled: Boolean, pushSettings: PushSettings): Result<Unit>
+    suspend fun updatePushSettings(enabled: Boolean, pushSettings: PushSettings?): Result<Unit>
```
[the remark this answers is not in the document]
antonijzelinskij: We can change settings before we disable it so if we will solve it using sealed class we should provide PushSettings to Disable state also but make it nullable.

I think it's not necessary: As a service it provides an API to domain/presentation layer that should control this state

feature-push-notifications/src/main/java/io/novafoundation/nova/feature_push_notifications/data/data/PushNotificationsService.kt
```diff
@@ -57,18 +67,35 @@ class RealPushNotificationsService(
...
         if (!googleApiAvailabilityProvider.isAvailable()) throw IllegalStateException("Google API is not available")
 
         return runCatching {
-            setPushNotificationsEnabled(enabled)
+            handlePushTokenIfNeeded(enabled)
             val pushToken = getPushToken()
             val oldSettings = settingsProvider.getPushSettings()
             subscriptionService.handleSubscription(enabled, pushToken, oldSettings, pushSettings)
+            settingsProvider.setPushNotificationsEnabled(enabled)
             settingsProvider.updateSettings(pushSettings)
```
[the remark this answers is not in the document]
antonijzelinskij: I also don’t see a problem with this. We have provided an API for use outside the class and manage the state externally as well

feature-push-notifications/src/main/java/io/novafoundation/nova/feature_push_notifications/data/domain/interactor/PushNotificationsInteractor.kt
antonijzelinskij: It seems to me that we should reset the push settings when we delete the last meta account in the application, so that when the user creates a new meta account and enters the push settings, he will see the default settings

feature-push-notifications/src/main/java/io/novafoundation/nova/feature_push_notifications/data/presentation/handling/NotificationHandlerExt.kt
antonijzelinskij: This flag indicating that the created PendingIntent should be immutable. This means that the additional intent argument passed to the send methods to fill in unpopulated properties of this intent will be ignored.

I passed it since high API requires MUTABLE or IMMUTABLE flag.

feature-push-notifications/src/main/java/io/novafoundation/nova/feature_push_notifications/data/data/PushNotificationsService.kt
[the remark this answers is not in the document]
valentunn: Ah you are right, we are storing settings even when notifications are disabled

feature-push-notifications/src/main/java/io/novafoundation/nova/feature_push_notifications/data/domain/interactor/PushNotificationsInteractor.kt
valentunn: So the initial logic was like that:
if notificationsEnabled && noAccounts - we do the first branch
else if noAccounts - we do the second branch

So if we are doing the second branch, the first one has failed => notificationsAreDisabled

I guess I've commented because I forgot that we're still storing notification settings even when notifications are disabled for easy re-enabling

valentunn: Here is what my thinking process looked like:
if `notificationsEnabled && noAccounts` - we do the first branch
else if `noAccounts` - we do the second branch

So if we are doing the second branch, the first one has failed => notificationsAreDisabled

I guess I've commented because I forgot that we're still storing notification settings even when notifications are disabled for easy re-enabling

feature-push-notifications/src/main/java/io/novafoundation/nova/feature_push_notifications/data/data/PushNotificationsService.kt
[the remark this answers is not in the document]
valentunn: Yeah again I forgot that notifications might be stored even when disabled

feature-push-notifications/src/main/java/io/novafoundation/nova/feature_push_notifications/data/domain/interactor/GovernancePushSettingsInteractor.kt
```diff
@@ -28,16 +29,17 @@ class RealGovernancePushSettingsInteractor(
...
             }
     }
 
-    private fun Chain.chainToGovTypes(): List<Pair<Chain, Chain.Governance>> {
-        return governance.filter { it == Chain.Governance.V2 }
-            .map { this to it }
+    private fun Chain.supportedGovTypes(): List<Pair<Chain, Chain.Governance>> {
+        return openGovIfSupported()?.let {
+            listOf(this to it)
+        }.orEmpty()
```
valentunn, line 42: I guess `listOfNonNull(openGovIfSupported()?.let { this to it })` is a lttile bit cleaner

runtime/src/main/java/io/novafoundation/nova/runtime/ext/ChainExt.kt
```diff
@@ -426,3 +426,7 @@ fun StatemineAssetId.onChainAssetId(): String {
         is StatemineAssetId.ScaleEncoded -> scaleHex
     }
 }
+
+fun Chain.openGovIfSupported(): Chain.Governance? {
+    return if (governance.contains(Chain.Governance.V2)) Chain.Governance.V2 else null
```
valentunn, line 431: `Chain.Governance.V2.takeIf { it in governance }`

valentunn (approved):

PR #1422, opened 2024-03-13
Disable push settings completely if other push settings is disabled

antonijzelinskij:
#86941rqbz

### DIFF
--- a/feature-push-notifications/src/main/java/io/novafoundation/nova/feature_push_notifications/data/presentation/handling/BaseNotificationHandler.kt
+++ b/feature-push-notifications/src/main/java/io/novafoundation/nova/feature_push_notifications/data/presentation/handling/BaseNotificationHandler.kt
@@ -9,7 +9,6 @@ import com.google.firebase.messaging.RemoteMessage
 import com.google.gson.Gson
 import io.novafoundation.nova.common.interfaces.ActivityIntentProvider
 import io.novafoundation.nova.common.resources.ResourceManager
-import io.novafoundation.nova.common.utils.filterNotNull
 import io.novafoundation.nova.common.utils.fromJson
 
 abstract class BaseNotificationHandler(
--- a/feature-push-notifications/src/main/java/io/novafoundation/nova/feature_push_notifications/data/presentation/handling/BaseNotificationHandler.kt
+++ b/feature-push-notifications/src/main/java/io/novafoundation/nova/feature_push_notifications/data/presentation/handling/BaseNotificationHandler.kt
@@ -9,6 +9,7 @@ import com.google.firebase.messaging.RemoteMessage
 import com.google.gson.Gson
 import io.novafoundation.nova.common.interfaces.ActivityIntentProvider
 import io.novafoundation.nova.common.resources.ResourceManager
+import io.novafoundation.nova.common.utils.filterNotNull
 import io.novafoundation.nova.common.utils.fromJson
 
 abstract class BaseNotificationHandler(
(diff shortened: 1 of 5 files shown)


stepanLav:
Task linked: [CU-86941rqbz Android. Disable notifications if all notification types are off](https://app.clickup.com/t/86941rqbz)

feature-push-notifications/src/main/java/io/novafoundation/nova/feature_push_notifications/data/presentation/handling/BaseNotificationHandler.kt
```diff
@@ -47,12 +48,12 @@ abstract class BaseNotificationHandler(
     protected abstract suspend fun handleNotificationInternal(channelId: String, message: RemoteMessage): Boolean
 
     internal fun RemoteMessage.getMessageContent(): NotificationData {
-        val payload: Map<String, Any> = data["payload"]?.let { payload -> gson.fromJson(payload) } ?: emptyMap()
+        val payload: Map<String, Any?> = data["payload"]?.let { payload -> gson.fromJson(payload) } ?: emptyMap()
```
valentunn, line 51: Thats why we had an issue with "assetId: null" right?

feature-push-notifications/src/main/java/io/novafoundation/nova/feature_push_notifications/data/presentation/settings/PushSettingsViewModel.kt
```diff
@@ -245,4 +250,20 @@ class PushSettingsViewModel(
...
     }
+
+    private fun disableNotificationsIfPushSettingsEmpty() {
+        pushSettingsState
+            .onEach { pushSettings ->
+                if (pushSettings?.settingsIsEmpty() == true) {
+                    pushEnabledState.value = false
+                }
+            }
+            .launchIn(this)
```
valentunn, line 261: I would rather suggest

``` kotlin
pushSettingsState
.filterNonNull()
.onEach { pushSettings ->
                if (pushSettings.settingsIsEmpty()) {
                    pushEnabledState.value = false
                }
```

To make `pushSettings == null` case more explicit

feature-push-notifications/src/main/java/io/novafoundation/nova/feature_push_notifications/data/presentation/settings/PushSettingsViewModel.kt
```diff
@@ -245,4 +250,20 @@ class PushSettingsViewModel(
...
+    }
+
+    private suspend fun makeDefaultPushSettingsIfNeeded() {
```
valentunn, line 264: I suggest `setDefaultPushSettingsIfEmpty()` - `ifNeeded` is rather generic and requires diving into the method's implementation to understand what "needed" means) I would use it only if we cant come up with more specific name, e.g. it is hard to reflect the behavior in the name

valentunn (approved):

feature-push-notifications/src/main/java/io/novafoundation/nova/feature_push_notifications/data/presentation/handling/BaseNotificationHandler.kt
antonijzelinskij: It seems that the main reason was the incorrect names of the notification types, and gson was already processing assetId: null.

I left this code just in case, but now it seems to me that it is superfluous

feature-push-notifications/src/main/java/io/novafoundation/nova/feature_push_notifications/data/presentation/settings/PushSettingsViewModel.kt
antonijzelinskij: Agree, fixed!

feature-push-notifications/src/main/java/io/novafoundation/nova/feature_push_notifications/data/presentation/handling/BaseNotificationHandler.kt
antonijzelinskij: Returned to the previous method

PR #1427, opened 2024-03-15
Fix/ref state and push sync

antonijzelinskij:


feature-push-notifications/src/main/java/io/novafoundation/nova/feature_push_notifications/data/PushNotificationsService.kt
```diff
@@ -28,7 +32,7 @@ interface PushNotificationsService {
 
     fun isPushNotificationsAvailable(): Boolean
 
-    suspend fun syncSettings()
+    suspend fun syncSettingsIfNeeded()
```
valentunn, line 35: I would suggest to create a separate method for periodic syncing, because we cant easily change the behavior of a method which is used in multiple places
For example, if we only sync in case certain amount of days passed but the client code expects us to do sync in a certain place, that's will introduce a bug

antonijzelinskij: This may not be the best idea, we wouldn't want the sync to run twice or more times if several conditions suddenly match

antonijzelinskij: But I changed a code a bit to make it more readable

valentunn (approved):

PR #1428, opened 2024-03-15
Improved select push wallets screen

antonijzelinskij:


feature-account-impl/src/main/java/io/novafoundation/nova/feature_account_impl/presentation/account/list/multipleSelecting/SelectMultipleWalletsViewModel.kt
```diff
@@ -40,6 +42,15 @@ class SelectMultipleWalletsViewModel(
...
     override val mode: AccountHolder.Mode = AccountHolder.Mode.SELECT_MULTIPLE
 
+    val confirmButtonState = selectedMetaAccounts.map { selectedMetaAccounts ->
+        if (selectedMetaAccounts.ids.isEmpty()) {
+            val disabledText = resourceManager.getQuantityString(R.plurals.multiple_wallets_selection_min_button_text, request.min, request.min)
+            DescriptiveButtonState.Disabled(disabledText)
+        } else {
+            DescriptiveButtonState.Enabled(resourceManager.getString(R.string.common_confirm))
+        }
+    }
```
valentunn, line 52: I think we should check for request.min and not for isEmpty()

valentunn (approved):

PR #1440, opened 2024-03-20
Fixed slippage message

antonijzelinskij:
#8693c4vfm
I checked this validation and it's useless since we are detecting when slippage not in range on slippage setup screen.
I'm not sure if it would be useful in future and I think we can remove it.

But as a fastest fix I just fixed a text message params


stepanLav:
Task linked: [CU-8693c4vfm Android. Pass params to error Slippage must be specified between %s and %s](https://app.clickup.com/t/8693c4vfm)

PR #1467, opened 2024-04-09
Cloud backup diffing

valentunn:
Todo:

Handle modified accounts during `applyBackupDiff`
Write integration test to check that the whole thing works


feature-account-impl/src/main/java/io/novafoundation/nova/feature_account_impl/data/cloudBackup/LocalAccountsCloudBackupFacade.kt
```diff
@@ -0,0 +1,408 @@
...
+                val chainAccountSecrets = chainAccountsSecretsByAccountId[chainAccountPublicInfo.accountId]
+
+                chainAccountSecrets?.toLocalSecrets()
+            }
+        ).filterNotNull()
+    }
+
+    private suspend fun getAllBackupableAccounts(): List<JoinedMetaAccountInfo> {
+        return accountDao.getMetaAccountsByStatus(MetaAccountLocal.Status.ACTIVE)
+            .filter { it.metaAccount.type != MetaAccountLocal.Type.PROXIED }
```
antonijzelinskij, line 212: It would be better to use `when` here with all types since if we add new account type compiler will request considering it here

feature-account-impl/src/main/java/io/novafoundation/nova/feature_account_impl/data/cloudBackup/LocalAccountsCloudBackupFacade.kt
```diff
@@ -0,0 +1,408 @@
...
+    override suspend fun applyBackupDiff(diff: CloudBackupDiff, cloudVersion: CloudBackup) {
+        val localChangesToApply = diff.localChanges
+        if (localChangesToApply.isEmpty()) return
+
+        val metaAccountsByUuid = getAllBackupableAccounts().associateBy { it.metaAccount.globallyUniqueId }
+
+        accountDao.withTransaction {
+            applyLocalRemoval(localChangesToApply.removed, metaAccountsByUuid)
+            applyLocalAddition(localChangesToApply.added, cloudVersion)
+            applyLocalModification(localChangesToApply.removed, cloudVersion)
```
antonijzelinskij, line 150: `localChangesToApply.removed` should be `modified`

feature-account-impl/src/main/java/io/novafoundation/nova/feature_account_impl/data/repository/addAccount/LocalAddMetaAccountRepository.kt
```diff
@@ -24,6 +22,6 @@ class LocalAddMetaAccountRepository(
     override suspend fun addAccountInternal(payload: Payload): AddAccountResult {
         val metaId = metaAccountDao.insertMetaAccount(payload.metaAccountLocal)
         secretStoreV2.putMetaAccountSecrets(metaId, payload.secrets)
-        return AddAccountResult.AccountAdded(metaId)
+        return AddAccountResult.AccountAdded(metaId, LightMetaAccount.Type.SECRETS)
```
antonijzelinskij, line 25: We should use type in payload.metaAccountLocal since we can't be sure it's always SECRET

antonijzelinskij (approved):

PR #1474, opened 2024-04-15
Fix Subsquare OpenGov API

valentunn:
* Only request 100 entries
* Only call Subsquare API once when screen is opened
* Do not wait for Subsquare  to return results and display referenda without metadata untill Subsquare results are loaded (to match iOS implementation)
#8694988u4

image


stepanLav:
Task linked: [CU-8694988u4 Android. Subsquare API fix](https://app.clickup.com/t/8694988u4)

antonijzelinskij (approved):

PR #1482, opened 2024-04-17
Change backup password flow

antonijzelinskij:
#8694997hj

### DIFF
--- a/feature-cloud-backup-api/src/main/java/io/novafoundation/nova/feature_cloud_backup_api/domain/CloudBackupService.kt
+++ b/feature-cloud-backup-api/src/main/java/io/novafoundation/nova/feature_cloud_backup_api/domain/CloudBackupService.kt
@@ -69,9 +69,9 @@ suspend fun CloudBackupService.fetchAndDecryptExistingBackup(password: String): 
  * @throws InvalidBackupPasswordError
  */
 suspend fun CloudBackupService.fetchAndDecryptExistingBackupWithSavedPassword(): Result<CloudBackup> {
-    return fetchBackup()
-        .flatMap { encryptedBackup ->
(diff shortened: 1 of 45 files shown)


stepanLav:
Task linked: [CU-8694997hj Android. Cloud backup menu actions — Change password](https://app.clickup.com/t/8694997hj)

feature-cloud-backup-api/src/main/java/io/novafoundation/nova/feature_cloud_backup_api/domain/CloudBackupService.kt
```diff
@@ -69,9 +69,9 @@ suspend fun CloudBackupService.fetchAndDecryptExistingBackup(password: String):
  * @throws InvalidBackupPasswordError
  */
 suspend fun CloudBackupService.fetchAndDecryptExistingBackupWithSavedPassword(): Result<CloudBackup> {
-    return session.getSavedPassword().flatMap { savedPassword ->
-        fetchBackup().flatMap { encryptedBackup ->
-            encryptedBackup.decrypt(savedPassword)
+    return fetchBackup()
+        .flatMap { encryptedBackup ->
+            val password = session.getSavedPassword().getOrThrow()
+            encryptedBackup.decrypt(password)
```
valentunn, line 75: Iirc `getOrThrow` in `flatMap` wont catch exeptions so it will be propagated beyond Result
What is the reason for such a change? Old code looks correct

antonijzelinskij: I fixed it to get an error "BackupNotFound" before "PasswordNotSet", since it's important for case when we enable backup sync by switcher

PR #1486, opened 2024-04-19
Cloud backup states

antonijzelinskij:
#8694998b9
#8694ahmzn
#8694998e0
#8694998tm


stepanLav:
Task linked: [CU-8694998b9 Android. Unsynced backup — Authentication failed](https://app.clickup.com/t/8694998b9)
Task linked: [CU-8694998e0 Android. Unsynced backup — Backup corrupted or empty (both cases)](https://app.clickup.com/t/8694998e0)
Task linked: [CU-8694998tm Android. Unsynced backup — Unknown error](https://app.clickup.com/t/8694998tm)
Task linked: [CU-8694ahmzn Android. Sync backup if not found](https://app.clickup.com/t/8694ahmzn)

feature-account-impl/src/main/java/io/novafoundation/nova/feature_account_impl/domain/startCreateWallet/StartCreateWalletInteractor.kt
```diff
@@ -15,4 +17,8 @@ class RealStartCreateWalletInteractor(
     override suspend fun validateCanCreateBackup(): PreCreateValidationStatus {
         return cloudBackupService.validateCanCreateBackup()
     }
+
+    override suspend fun signInToCloud(): Result<Unit> {
```
valentunn, line 21: There is no point doing that for every screen since cloud backup service ensures that user is authenticated for every action in the service

feature-settings-impl/src/main/java/io/novafoundation/nova/feature_settings_impl/domain/CloudBackupSettingsInteractor.kt
```diff
@@ -38,7 +43,7 @@ class RealCloudBackupSettingsInteractor(
                 cloudBackupFacade.applyNonDestructiveCloudVersionOrThrow(cloudBackup, BackupDiffStrategy.syncWithCloud())
 
                 Unit
-            }.finally {
+            }.onSuccess {
```
valentunn, line 46: I think it should be `finally` - we probably want to display updated timestamp even on failed sync

feature-account-impl/src/main/java/io/novafoundation/nova/feature_account_impl/domain/startCreateWallet/StartCreateWalletInteractor.kt
antonijzelinskij: I added this function since we have an action "Sign in" in design and as we discussed this button should start authentication
image

antonijzelinskij: According to design - yes, we have this dialogue with the same action for every screen I added

valentunn: But according to the code there is no dialogue for other screens, it is done in view model init instead. That's why I am saying it is redundant - storage does the same automatically when doing any action

valentunn (approved):

PR #1490, opened 2024-04-27
Fetch Nova validators from remote

valentunn:
#8694dpefr

PR #1491, opened 2024-04-27
Fetch Nova validators from remote

valentunn:
#8694dpefr

Closes #1490 and does the same but for develop branch


stepanLav:
Task linked: [CU-8694dpefr Android. Update Recommended validators list to use remote config](https://app.clickup.com/t/8694dpefr)

antonijzelinskij (approved):

feature-staking-impl/src/main/java/io/novafoundation/nova/feature_staking_impl/data/validators/NovaValidatorsApi.kt
```diff
@@ -0,0 +1,9 @@
+package io.novafoundation.nova.feature_staking_impl.data.validators
+
+import retrofit2.http.GET
+
+interface NovaValidatorsApi {
+
+    @GET("https://raw.githubusercontent.com/novasamatech/nova-utils/master/staking/nova_validators.json")
```
ERussel, line 7: Probably we also need a separate config for dev in case we want to check how recommended validators list will look like without risking users can select those validators

ERussel (approved):

feature-staking-impl/src/main/java/io/novafoundation/nova/feature_staking_impl/data/validators/KnownNovaValidators.kt
```diff
@@ -1,41 +1,61 @@
...
+    override suspend fun getValidatorIds(chainId: ChainId): List<String> {
+        return getValidators()[chainId].orEmpty()
     }
 
-    override fun getValidatorIds(chainId: ChainId): Set<String> {
-        return novaValidators[chainId].orEmpty()
+    private suspend fun getValidators(): Map<String, List<String>> {
+        return validatorsMutex.withLock {
+            if (validatorsByNetwork == null) {
+                validatorsByNetwork = fetchValidators()
```
ERussel, line 33: Probably, we don't want to cache empty list in case request is failed

valentunn: Thanks for spotting this. I have concern that not caching failed result might lead potential inconsistencies in recommendations within one session. But your suggestion is valid, I will think how can we improve this

PR #1497, opened 2024-05-01
Select Account to backup

antonijzelinskij:
#8694dmmct

### DIFF
--- a/feature-account-impl/src/main/java/io/novafoundation/nova/feature_account_impl/domain/manualBackup/ManualBackupInteractor.kt
+++ b/feature-account-impl/src/main/java/io/novafoundation/nova/feature_account_impl/domain/manualBackup/ManualBackupInteractor.kt
@@ -8,18 +8,53 @@ import io.novafoundation.nova.feature_account_api.domain.model.LightMetaAccount.
 import io.novafoundation.nova.feature_account_api.domain.model.LightMetaAccount.Type.POLKADOT_VAULT
 import io.novafoundation.nova.feature_account_api.domain.model.LightMetaAccount.Type.PROXIED
 import io.novafoundation.nova.feature_account_api.domain.model.MetaAccount
+import io.novafoundation.nova.feature_account_api.domain.model.hasAccountIn
+import io.novafoundation.nova.runtime.ext.defaultComparatorFrom
+import io.novafoundation.nova.runtime.multiNetwork.ChainRegistry
+import io.novafoundation.nova.runtime.multiNetwork.chain.model.Chain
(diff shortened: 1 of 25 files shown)


stepanLav:
Task linked: [CU-8694dmmct Android. Backup manually — Select key to backup](https://app.clickup.com/t/8694dmmct)

feature-account-impl/src/main/java/io/novafoundation/nova/feature_account_impl/domain/manualBackup/ManualBackupInteractor.kt
```diff
@@ -8,20 +8,55 @@ import io.novafoundation.nova.feature_account_api.domain.model.LightMetaAccount.
...
+import kotlinx.coroutines.flow.Flow
+import kotlinx.coroutines.flow.map
+
+class MetaAccountChains(
+    val metaAccount: MetaAccount,
+    val defaultChains: List<Chain>,
+    val customChains: List<Chain>
+)
 
 interface ManualBackupInteractor {
```
valentunn, line 24: Not sure we need to put everything into single interactor. I suggest to create one interactor per screen

feature-account-impl/src/main/java/io/novafoundation/nova/feature_account_impl/presentation/manualBackup/accounts/ManualBackupSelectAccountViewModel.kt
```diff
@@ -0,0 +1,81 @@
...
+
+    fun backClicked() {
+        router.back()
+    }
+
+    private fun mapMetaAccountToUI(
+        metaAccountChains: MetaAccountChains
+    ): List<ManualBackupRVItem> {
+        return buildList {
+            if (metaAccountChains.defaultChains.isNotEmpty()) {
```
valentunn, line 47: Iirc we have similar logic in Wallet Connect Session Approve screen. Can you please check whether we can reuse it or not?

valentunn (approved):

feature-account-impl/src/main/java/io/novafoundation/nova/feature_account_impl/presentation/manualBackup/accounts/ManualBackupSelectAccountViewModel.kt
antonijzelinskij: I checked it and think this logic is not very reusable in our case and also my logic is pretty simple and I think we can left it like that

valentunn (approved):

PR #1499, opened 2024-05-02
Manual Backup Warning Fragment

antonijzelinskij:
#8694dmmq2


stepanLav:
Task linked: [CU-8694dmmq2 Android. Backup manually — Warnings screen with Biometry/PIN requirement](https://app.clickup.com/t/8694dmmq2)

common/src/main/java/io/novafoundation/nova/common/mixin/condition/ConditionMixin.kt
```diff
@@ -0,0 +1,23 @@
...
+    fun checkCondition(index: Int, isChecked: Boolean)
+}
+
+interface ConditionMixinFactory {
+
+    fun createConditionMixin(
+        coroutineScope: CoroutineScope,
+        conditionsCount: Int,
+        @StringRes enabledButtonText: Int,
+        @StringRes disabledButtonText: Int,
```
valentunn, line 21: Do we really need to make butten texts a part of mixin implementation? Imho we can keep interface minimal and just have `allConditionsSatisfied: Flow<Boolean>` there. Then, we may have an extension `buttonState(enabledText, disabledText): Flow< DescriptiveButtonState >`.This way we
1. Better conform to single responsibility
2. Dont force clients of ConditionMixin to always setup a button

common/src/main/java/io/novafoundation/nova/common/view/WarningCheckBox.kt
```diff
@@ -0,0 +1,58 @@
...
+import io.novafoundation.nova.common.utils.setImageTint
+import io.novafoundation.nova.common.utils.useAttributes
+import kotlinx.android.synthetic.main.view_warning_checkbox.view.warningCheckBoxCheckBox
+import kotlinx.android.synthetic.main.view_warning_checkbox.view.warningCheckBoxIcon
+
+class WarningCheckBox @JvmOverloads constructor(
+    context: Context,
+    attrs: AttributeSet? = null,
+    defStyleAttr: Int = 0
+) : LinearLayout(context, attrs, defStyleAttr) {
```
valentunn, line 21: We can implement `Checkable` here, see the reason in the next comment

feature-account-impl/src/main/java/io/novafoundation/nova/feature_account_impl/presentation/manualBackup/warning/ManualBackupWarningViewModel.kt
```diff
@@ -0,0 +1,34 @@
...
+import androidx.lifecycle.viewModelScope
+import io.novafoundation.nova.common.base.BaseViewModel
+import io.novafoundation.nova.common.mixin.condition.ConditionMixinFactory
+import io.novafoundation.nova.feature_account_impl.R
+import io.novafoundation.nova.feature_account_impl.presentation.AccountRouter
+import io.novafoundation.nova.feature_account_impl.presentation.manualBackup.common.ManualBackupAccountToBackupPayload
+
+const val CONDITION_ID_1 = 0
+const val CONDITION_ID_2 = 1
+const val CONDITION_ID_3 = 2
```
valentunn, line 12: This seems like a part that will be duplicated each time we use `ConditionsMixin`
My suggestion is to have an extension method

`ConditionsMixin.setupConditions(varag conditionInputs: Checkable)` for usage in the ui

valentunn (approved):

feature-account-impl/src/main/java/io/novafoundation/nova/feature_account_impl/presentation/manualBackup/warning/ManualBackupWarningViewModel.kt
antonijzelinskij: Nice solution!

But it's not enough to implement Checkable interface since it doesn't have observe method. Added `CheckableListener `for this case

valentunn (approved):

PR #1510, opened 2024-05-10
Refactor export json fragment

antonijzelinskij:
#8694gg39a


stepanLav:
Task linked: [CU-8694gg39a Android. Redesign Export Json screen](https://app.clickup.com/t/8694gg39a)

feature-account-api/src/main/java/io/novafoundation/nova/feature_account_api/domain/model/MetaAccount.kt
```diff
@@ -154,6 +154,10 @@ fun MetaAccount.publicKeyIn(chain: Chain): ByteArray? {
     }
 }
 
+fun MetaAccount.multiChainEncryption(): MultiChainEncryption? {
```
valentunn, line 157: The name sounds confusing - maybe better to name it something like `substrateMultiChainEncryption`?

feature-account-impl/src/main/java/io/novafoundation/nova/feature_account_impl/data/repository/AccountRepositoryImpl.kt
```diff
@@ -260,6 +262,25 @@ class AccountRepositoryImpl(
...
+        return withContext(Dispatchers.Default) {
+            val secrets = secretStoreV2.getMetaAccountSecrets(metaAccount.id)!!
+
+            jsonSeedEncoder.generate(
+                keypair = secrets.keypair(ethereum = false),
+                seed = secrets.seed,
+                password = password,
+                name = metaAccount.name,
+                multiChainEncryption = metaAccount.multiChainEncryption()!!,
+                genesisHash = "",
```
valentunn, line 278: I dont think we can leave the genesis hash empty - can you check whether such JSON is compatible with Import to Nova and Import to Polkadot.js?

antonijzelinskij: Thank you for advice!

It's compatable. I also checked that JSON imported from polkadot.js also doesn't have genesisHash

valentunn (approved):

PR #1518, opened 2024-05-16
Ledger usb transport

valentunn:


valentunn:
Tested - both getting the address and signing (multiple chunks per command) works

antonijzelinskij (approved):

PR #1536, opened 2024-05-27
Generic ledger to dev

valentunn:
Diff to check: https://github.com/novasamatech/nova-wallet-android/pull/1536/files/eef38e5489f29163a98730be00e05ca077c7b806..1dde6d56f14195b929694e800e39ff3c50eebcff

Notable changes:

1. Integrate LEDGER_GENERIC into cloud backup feature: we record the universal DP as a DP of a single chain account secrets. Add tests for it
2. Adjust AddAccountRepository to support updating multiple accounts if payload dictates so. Use it to perform batch update for missing generic ledger chains so cloud backup write will be triggered only once


feature-account-api/src/main/java/io/novafoundation/nova/feature_account_api/data/repository/addAccount/AddAccountRepository.kt
```diff
@@ -9,13 +9,35 @@ interface AddAccountRepository<T> {
...
 
-    class AccountChanged(override val metaId: Long) : AddAccountResult
+    class AccountAdded(override val metaId: Long, val type: LightMetaAccount.Type) : HadEffect, SingleAccountChange
 
-    class NoOp(override val metaId: Long) : AddAccountResult
+    class AccountChanged(override val metaId: Long, val type: LightMetaAccount.Type) : HadEffect, SingleAccountChange
+
+    class Batch(val updates: List<HadEffect>) : HadEffect
+
+    object NoOp : AddAccountResult
```
antonijzelinskij, line 25: I think it would be better to name it without abbreviation

antonijzelinskij (approved):

feature-account-api/src/main/java/io/novafoundation/nova/feature_account_api/data/repository/addAccount/AddAccountRepository.kt
valentunn: It is quite a standard and well-known term: https://ru.wikipedia.org/wiki/NOP. I think its fine having it like that especially considering that a lot of codebases do the same (e.g. Substrate does)

PR #1541, opened 2024-05-30
Added networks

antonijzelinskij:
#8694k6c1h
#8694k6c16


stepanLav:
Task linked: [CU-8694k6c16 Android. Networks screen. Added tab](https://app.clickup.com/t/8694k6c16)
Task linked: [CU-8694k6c1h Android. Update database chain entity to support add/remove enable/disable network](https://app.clickup.com/t/8694k6c1h)

common/src/main/java/io/novafoundation/nova/common/utils/ViewSpace.kt
```diff
@@ -0,0 +1,31 @@
+package io.novafoundation.nova.common.utils
+
+import android.view.View
+import android.view.ViewGroup
+
```
valentunn, line 5: Sounds very similar to Rect or RectF

core-db/src/main/java/io/novafoundation/nova/core_db/migrations/58_59_ChainNetworkManagement.kt
```diff
@@ -0,0 +1,12 @@
+package io.novafoundation.nova.core_db.migrations
+
+import androidx.room.migration.Migration
+import androidx.sqlite.db.SupportSQLiteDatabase
+import io.novafoundation.nova.core_db.model.chain.ChainLocal
+
+val ChainNetworkManagement_58_59 = object : Migration(58, 59) {
+
+    override fun migrate(database: SupportSQLiteDatabase) {
+        database.execSQL("ALTER TABLE chains ADD COLUMN `isCustomNetwork` INTEGER NOT NULL DEFAULT ${ChainLocal.DEFAULT_NETWORK_BY_DEFAULT_STR}")
```
valentunn, line 10: I suggest to make it an enum, similarilly to what we have in the Chain.Asset: `enum Source { DEFAULT, CUSTOM }`. This might prevent us to not to write migrations in future if we will add new types of sources

feature-settings-impl/src/main/java/io/novafoundation/nova/feature_settings_impl/domain/NetworkManagementInteractor.kt
```diff
@@ -8,29 +10,61 @@ import io.novasama.substrate_sdk_android.wsrpc.state.SocketStateMachine
...
+                connectionsFlow(sortChains(chains))
+            }
+    }
+
+    override fun addedNetworksFlow(): Flow<List<NetworkState>> {
+        return chainRegistry.currentChains
+            .filterList { it.isCustomNetwork }
+            .flatMapLatest { chains ->
+                connectionsFlow(sortChains(chains))
+            }
```
valentunn, line 60: Those seems to be the same exept filtering condition. We might extract it:

```kotlin
fun networksFlow(filter: (Chain) -> Boolean) = chainRegistry.currentChains
            .filterList(filter)
            .flatMapLatest { chains ->
                connectionsFlow(sortChains(chains))
            }

File: runtime/src/main/java/io/novafoundation/nova/runtime/multiNetwork/ChainRegistry.kt [hunk 4]
```diff
@@ -320,3 +329,12 @@ suspend fun ChainRegistry.findEvmChainFromHexId(evmChainIdHex: String): Chain? {
 
     return findEvmChain(addressPrefix)
 }
+
+fun ChainRegistry.enabledChains() = currentChains
+    .filterList { it.isEnabled }
+    .inBackground()
+    .shareIn(this, SharingStarted.Eagerly, replay = 1)
```
valentunn, line 336: I dont think we should use sharing here:

1. The sharing is done in `ChainRegistry` scope which is basically global one. So you will leak each flow you request this way and all internal information (chains map)
2. This is quite an unexpected behavior from the **function**. Usually we done sharing for flows stored in the class fields. Sharing of a flow returned from a function call wont make any difference unless you explicitly assign it to a variabler and resuse multiple times. Which is not how you use it (as I can see)

runtime/src/main/java/io/novafoundation/nova/runtime/multiNetwork/ChainRegistry.kt
```diff
@@ -320,3 +329,12 @@ suspend fun ChainRegistry.findEvmChainFromHexId(evmChainIdHex: String): Chain? {
...
+
+fun ChainRegistry.enabledChainById() = enabledChains().map { chains -> chains.associateBy { it.id } }
+    .inBackground()
+    .shareIn(this, SharingStarted.Eagerly, replay = 1)
```
valentunn, line 340: Same here

runtime/src/main/java/io/novafoundation/nova/runtime/multiNetwork/chain/ChainSyncService.kt
```diff
@@ -37,7 +37,7 @@ class ChainSyncService(
 
         val remoteChains = retryUntilDone { chainFetcher.getChains() }
 
-        val newChains = remoteChains.map { mapRemoteChainToLocal(it, oldChainsById[it.chainId], gson) }
+        val newChains = remoteChains.map { mapRemoteChainToLocal(it, oldChainsById[it.chainId], isCustomNetwork = false, gson) }
```
valentunn, line 40: It is not an issue in the scope of this PR, but once weimplement "add network" feature we will need to adapt chain diffing here in ChainSyncService since otherwise it will remove manually added chains that are not present in chains.json upon each sync

feature-wallet-api/src/main/java/io/novafoundation/nova/feature_wallet_api/domain/implementations/SelectableAssetUseCaseImpl.kt
```diff
@@ -26,11 +27,13 @@ class SelectableAssetUseCaseImpl<A : SelectableAssetAdditionalData>(
 
         val balancesByChainAssets = walletRepository.getSupportedAssets(metaAccount.id).associateBy { it.token.configuration.fullId }
 
-        sharedState.availableToSelect().mapNotNull { supportedOption ->
-            val asset = balancesByChainAssets[supportedOption.assetWithChain.asset.fullId]
+        sharedState.availableToSelect()
```
valentunn, line 30: I think it is better to modify `availableToSelect`

common/src/main/java/io/novafoundation/nova/common/utils/ViewSpace.kt
antonijzelinskij: Yes but it's from android graphic library. I don't want to mix it

runtime/src/main/java/io/novafoundation/nova/runtime/multiNetwork/chain/ChainSyncService.kt
antonijzelinskij: Agree. I think we can handle chains here by source

valentunn (approved):

PR #1559, opened 2024-06-26
Rc/8.0.1

valentunn:
I didnt figured out why exactly does `UPDATE meta_accounts AS m`  trigger  `synatx error near "as"` error on older version of Android (no mentions of something similar in SQLite changelogs) but removing aliasing resolved the issue on my Android 10 emulator

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -1,8 +1,8 @@
 buildscript {
     ext {
         // App version
-        versionName = '8.0.0'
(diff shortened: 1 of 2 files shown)


valentunn:
Release severity: Normal
Release notes:
Fixed startup crash on Android 10 and older

antonijzelinskij (approved):

PR #1560, opened 2024-06-26
Create custom node

antonijzelinskij:
#8694k6c2h

### DIFF
--- a/common/src/main/java/io/novafoundation/nova/common/utils/Urls.kt
+++ b/common/src/main/java/io/novafoundation/nova/common/utils/Urls.kt
@@ -1,6 +1,7 @@
 package io.novafoundation.nova.common.utils
 
 import android.util.Patterns
+import java.net.URI
 import java.net.URL
 
 object Urls {
@@ -15,6 +16,12 @@ object Urls {
         val parsedUrl = URL(url)
 
         return "${parsedUrl.protocol}://${parsedUrl.host}"
+    }
+
+    fun normalizePath(url: String): String {
+        val path = if (url.endsWith('/')) url.substring(0, url.length - 1) else url
+
+        return URI.create(path).normalize().toString()
     }
 
     fun hostOf(url: String): String {
--- a/feature-settings-impl/src/main/java/io/novafoundation/nova/feature_settings_impl/data/NodeChainIdRepository.kt
+++ b/feature-settings-impl/src/main/java/io/novafoundation/nova/feature_settings_impl/data/NodeChainIdRepository.kt
@@ -1,0 +1,68 @@
+package io.novafoundation.nova.feature_settings_impl.data
+
+import io.novafoundation.nova.caip.caip2.identifier.Caip2Identifier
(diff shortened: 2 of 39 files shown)


stepanLav:
Task linked: [CU-8694k6c2h Android. Add custom network node screen](https://app.clickup.com/t/8694k6c2h)

common/src/main/java/io/novafoundation/nova/common/utils/Urls.kt
```diff
@@ -17,6 +18,12 @@ object Urls {
         return "${parsedUrl.protocol}://${parsedUrl.host}"
     }
 
+    fun normalizePath(url: String): String {
+        val path = if (url.endsWith('/')) url.substring(0, url.length - 1) else url
```
valentunn, line 22: Iirc there is an extension called `String.removeSuffix` to achieve exactly that

feature-settings-impl/src/main/java/io/novafoundation/nova/feature_settings_impl/data/NodeChainIdRepository.kt
```diff
@@ -0,0 +1,68 @@
...
+    suspend fun requestChainId(): String
+}
+
+class NodeChainIdRepositoryFactory(
+    private val nodeConnectionFactory: NodeConnectionFactory,
+    private val web3ApiFactory: Web3ApiFactory
+) {
+
+    fun create(chain: Chain, nodeUrl: String): NodeChainIdRepository {
+        val nodeConnection = nodeConnectionFactory.createNodeConnection(nodeUrl)
```
valentunn, line 28: We should make sure the connection is shut down once we done validating it

feature-settings-impl/src/main/java/io/novafoundation/nova/feature_settings_impl/domain/CustomNodeInteractor.kt
```diff
@@ -0,0 +1,52 @@
...
+import io.novafoundation.nova.runtime.multiNetwork.chain.model.Chain
+import io.novafoundation.nova.runtime.repository.ChainNodeRepository
+
+interface CustomNodeInteractor {
+
+    suspend fun getNodeDetails(chainId: String, nodeUrl: String): Result<Chain.Node>
+
+    suspend fun createNode(chainId: String, url: String, name: String)
+
+    suspend fun saveNode(chainId: String, oldUrl: String, url: String, name: String)
```
valentunn, line 18: I suggest to rename it to `updateNode`, otherwise it looks a bit confusing since we have `createNode` and `saveNode`

valentunn (approved):

feature-settings-impl/src/main/java/io/novafoundation/nova/feature_settings_impl/data/NodeChainIdRepository.kt
antonijzelinskij: I think it will be nice to provide here coroutine scope and stop socket when coroutine scope is completed

PR #1561, opened 2024-06-27
Feat/adding_instrumentail_tests

stepanLav:


.github/scripts/run_instrumental_tests.sh
```diff
@@ -0,0 +1,52 @@
...
+    time.sleep(5)
+    print ("Running...")
+t = threading.Thread(target=update)
+t.dameon = True
+t.start()
+def run():
+  os.system('adb wait-for-device')
+  p = sp.Popen('adb shell am instrument -w -m -e notClass io.novafoundation.nova.balances.BalancesIntegrationTest -e package io.novafoundation.nova.debug io.novafoundation.nova.debug.test/io.qameta.allure.android.runners.AllureAndroidJUnitRunner',
+               shell=True, stdout=sp.PIPE, stderr=sp.PIPE, stdin=sp.PIPE)
+  return p.communicate()
```
antonijzelinskij, line 33: As I see we start only BalancesIntegrationTest. Is it correct?

.github/scripts/run_balances_test.sh
antonijzelinskij, line 1: Why we need to run balances test separately?
Is it because this test is moved to another directory?

app/src/androidTest/java/io/novafoundation/nova/MoonbaseSendIntagrationTest.kt
```diff
@@ -50,13 +55,37 @@ class MoonbaseSendIntagrationTest {
         return EthereumKeypairFactory.generate(seed, junctions)
     }
 
+    private inner class TestSigner(
```
antonijzelinskij, line 58: Nice solution I guess! But let's move inner class to the bottom of its parent

app/src/androidTest/java/io/novafoundation/nova/MoonbaseSendIntagrationTest.kt
```diff
@@ -50,13 +55,37 @@ class MoonbaseSendIntagrationTest {
...
+        private val testPayload: (SignerPayloadExtrinsic) -> Unit
```
antonijzelinskij, line 59: As I see we don't use testPayload from outside, so we can remove this redundant variable

app/src/androidTest/java/io/novafoundation/nova/MoonbaseSendIntagrationTest.kt
```diff
@@ -50,13 +55,37 @@ class MoonbaseSendIntagrationTest {
...
+    ) : NovaSigner {
+
+        val accountId = ByteArray(32) { 1 }
```
antonijzelinskij, line 62: It seems we don't use it but to be semantically correct let's provide accountId to the constructor from outside

MehrdadBadriGol (approved):

PR #1587, opened 2024-07-16
Fix/nomination pools

valentunn:
Nomination pools will soon migrate from transferring user funds to a pool account to locking funds inside user account which will provide user more opportunities to use those funds in another way as well
To support that, we implement the following:
* Each pool action (bond more, unbond, redeem, claim rewards) additinally executes migration call, if needed
* Each pool action + start staking have a validation that prevents doing the action if user has Direct Staking (with new logic it is not possible anymore to have direct and pool staking at the same time)
* When syncing pool external balance, we now reduce it by the amount that is reserved in the user account (obtained via delegatedStaking.delegator)
* PR also adds support for Holds (aka labeled reserves) to show the user pool amount that is now reserved on his account in the balance breakdown


feature-wallet-api/src/main/java/io/novafoundation/nova/feature_wallet_api/domain/model/BalanceHold.kt
```diff
@@ -0,0 +1,35 @@
...
+fun mapBalanceHoldFromLocal(
+    asset: Chain.Asset,
+    hold: BalanceHoldLocal
+): BalanceHold {
+    return BalanceHold(
+        id = hold.id.toDomain(),
+        amountInPlanks = hold.amount,
+        chainAsset = asset
+    )
+}
```
antonijzelinskij, line 31: It's not a best practice to keep data -> domain mappers in domain layer. Let's move it to the data directory?

antonijzelinskij (approved):

PR #1600, opened 2024-07-20
Rc/8.2.1

antonijzelinskij:


antonijzelinskij:
Release severity: Normal
Release notes:
Support for new runtime upgrades and security improvements are here in Nova Wallet v8.2.1

– Added support for new runtime upgrade affecting nomination pools
– Added security improvements for the DApp browser – Nova now rejects access to certain top level domains and warns users to be careful when connecting to DApps which are not already in the DApp catalog
– Fixes & Optimizations

PR #1606, opened 2024-07-22
Rc/8.3.0

antonijzelinskij:


antonijzelinskij:
Release severity: Normal
Release notes:
Meet the Avail Staking update — Stake AVAIL and get Maximized rewards in Nova Wallet! 🏆
— Added Avail 2-in-1 Simple Staking: Nova makes sure you start staking in the best possible way (validator or pool staking) for your desired AVAIL amount
— Added Avail Balances, Transfers, Operation history
— Added AVAIL Staking analytics: Total rewards, APY, Rewards history, Status
— Added DApp support for the Avail network

feature-staking-impl/src/main/java/io/novafoundation/nova/feature_staking_impl/domain/validators/ValidatorProvider.kt
```diff
@@ -91,7 +91,10 @@ class ValidatorProvider(
 
         val slashes = stakingRepository.getSlashes(chainId, accountIdBridged)
 
-        val novaValidatorIds = knownNovaValidators.getValidatorIds(chainId).toSet()
+        val novaValidatorIds = validatorsPreferencesSource.getRecommendedValidatorIds(chainId)
```
ERussel, line 94: I don't think we need to apply exclusion logic when we are getting a validator. I think the cause we are using preferred one is due to the validator might not be elected

antonijzelinskij: Agree, thank you!

ERussel (commented):
I don't get why we are exposing exclusion logic to the function that gets not elected validator

ERussel (approved):

PR #1607, opened 2024-07-23
Feature/abstain votes

antonijzelinskij:


feature-governance-impl/src/main/java/io/novafoundation/nova/feature_governance_impl/domain/referendum/vote/validations/VoteReferendumValidationPayload.kt
```diff
@@ -1,15 +1,19 @@
...
-class VoteReferendumValidationPayload(
+data class VoteReferendumValidationPayload(
     val onChainReferendum: OnChainReferendum,
     val asset: Asset,
     val trackVoting: Voting?,
     val voteAmount: BigDecimal,
-    val fee: DecimalFee
+    val fee: DecimalFee,
+    val voteType: VoteType,
+    val conviction: Conviction
```
ERussel, line 18: As conviction for abstain doesn't make sense I would suggest to have the conviction as associated value of the VoteType just to not having mismatching between voteType and conviction

ERussel (approved):

feature-governance-impl/src/main/java/io/novafoundation/nova/feature_governance_impl/domain/referendum/vote/validations/VoteReferendumValidationPayload.kt
antonijzelinskij: Since this is validation payload it's important to pass conviction here to check that Conviction is NONE for abstain vote and show a dialog.
But this is a valid point for confirmation screen payload. Thank you for this advice.

ERussel (approved):

PR #1630, opened 2024-08-08
Feature/custom fee logic

antonijzelinskij:
#86959176v


stepanLav:
Task linked: [CU-86959176v Android. Transfers. Add signed extension for Fees in custom asset (Asset Hub)](https://app.clickup.com/t/86959176v)

feature-account-impl/src/main/java/io/novafoundation/nova/feature_account_impl/data/fee/types/HydrationConversionFeePayment.kt
```diff
@@ -0,0 +1,57 @@
...
+
+    override suspend fun availableCustomFeeAssets(): List<Chain.Asset> {
+        val metaAccount = accountRepository.getSelectedMetaAccount()
+        val chain = chainRegistry.getChain(paymentAsset.chainId)
+        val accountId = metaAccount.accountIdIn(chain)
+        val fromAsset = chain.commissionAsset
+
+        val allSwapDirections = hydraDxQuoteSharedComputation.directions(chain, accountId!!, coroutineScope)
+        val commissionAssetDirections = allSwapDirections.adjacencyList[fromAsset.fullId] ?: emptyList()
+        return commissionAssetDirections.map { chainRegistry.asset(it.direction.to) }
```
valentunn, line 55: Are we sure this will handle a case when there is no direct route between HDX and fee token but there is an inderct route for it? E.g. HDX and USDT:
There is no direct HDX -> USDT route, but there is HDX -> StablePool -> USDT one

I cant recall whether adjencyList returned from hydra swap source includes only direct connections or all possible ones. I feel like the latter should be the case but its better to check

feature-account-impl/src/main/java/io/novafoundation/nova/feature_account_impl/data/fee/utils/HydraDxQuoteSharedComputation.kt
```diff
@@ -0,0 +1,76 @@
...
+        chain: Chain,
+        accountId: AccountId,
+        scope: CoroutineScope
+    ): AssetConversion<HydraDxSwapEdge> {
+        val key = "HydraDxAssetConversion:${accountId.toHexString()}"
+
+        return computationalCache.useCache(key, scope) {
+            val storageSharedRequestBuilder = storageSharedRequestsBuilderFactory.create(chain.id)
+            val assetConversion = assetConversionFactory.create(chain)
+            assetConversion.runSubscriptions(accountId, storageSharedRequestBuilder)
```
valentunn, line 72: Have you tested that subscriptions actually work? We never call `storageSharedRequestBuilder.subscribe` so seems like dont actually subscribe atm. We can subscribe based on AggregateScope that CompotationalCache uses. In that case we pass it to the lambda in `useCache` in case its not already passed there

feature-account-impl/src/main/java/io/novafoundation/nova/feature_account_impl/data/fee/utils/HydraDxQuoteSharedComputation.kt
```diff
@@ -0,0 +1,76 @@
...
+    suspend fun quote(
+        chain: Chain,
+        accountId: AccountId,
+        fromAsset: Chain.Asset,
+        toAsset: Chain.Asset,
+        scope: CoroutineScope
+    ): AssetExchangeQuote {
+        val key = "HydraDxQuote:${accountId.toHexString()}"
+
+        return computationalCache.useCache(key, scope) {
```
valentunn, line 47: This  is the tricky one - quotes may change based on onchain state so we should not cache `assetConversion.quote(paths, args)` if we want to be accurate

feature-swap-core/src/test/java/io/novafoundation/nova/ExampleUnitTest.kt
```diff
@@ -0,0 +1,17 @@
...
+import org.junit.Test
+
+import org.junit.Assert.*
+
+/**
+ * Example local unit test, which will execute on the development machine (host).
+ *
+ * See [testing documentation](http://d.android.com/tools/testing).
+ */
+class ExampleUnitTest {
```
valentunn, line 12: Can be deleted

runtime/src/main/java/io/novafoundation/nova/runtime/util/FullAssetIdModel.kt
```diff
@@ -0,0 +1,47 @@
+package io.novafoundation.nova.runtime.util
+
+import android.os.Parcel
+import android.os.Parcelable
+import io.novafoundation.nova.runtime.multiNetwork.chain.model.ChainId
+import io.novafoundation.nova.runtime.multiNetwork.chain.model.FullChainAssetId
+
+class FullAssetIdModel(val chainId: ChainId, val assetId: Int) : Parcelable {
+
+    constructor(parcel: Parcel) : this(
```
valentunn, line 10: Cant we use Parcelize here?

feature-wallet-impl/src/main/java/io/novafoundation/nova/feature_wallet_impl/data/network/blockchain/assets/transfers/BaseAssetTransfers.kt
```diff
@@ -48,16 +48,20 @@ abstract class BaseAssetTransfers(
...
     }
 
-    override suspend fun calculateFee(transfer: AssetTransfer): Fee {
-        return extrinsicService.estimateFee(transfer.originChain, TransactionOrigin.SelectedWallet) {
-            transfer(transfer)
-        }
+    override suspend fun calculateFee(transfer: AssetTransfer, coroutineScope: CoroutineScope): Fee {
+        return extrinsicServiceFactory
+            .create(coroutineScope)
+            .estimateFee(transfer.originChain, TransactionOrigin.SelectedWallet) {
```
valentunn, line 62: Can you please point me where do we pass FeePayment.Asset based on the `transfer`? Also I havent seen a modification of FeeLoaderMixin to account for fee selector, is it going to be a separate PR?

valentunn (commented):
Overall approach looks good, we just need to sort of comments related to correctness below

feature-account-api/src/main/java/io/novafoundation/nova/feature_account_api/data/model/Fee.kt
```diff
@@ -23,10 +24,15 @@ data class EvmFee(
     override val amount = gasLimit * gasPrice
 }
 
+interface FeeInAsset : Fee {
+    val assetId: FullChainAssetId
```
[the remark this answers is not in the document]
antonijzelinskij: I used ids only because it's not very easy to pass Fee using fragment payload (SimpleFeeParcelModel) if we have ChainAsset model inside.

feature-account-impl/src/main/java/io/novafoundation/nova/feature_account_impl/data/fee/utils/HydraDxQuoteSharedComputation.kt
antonijzelinskij: Yes it doesn't work. I fixed it in [this](https://github.com/novasamatech/nova-wallet-android/pull/1633) pr

runtime/src/main/java/io/novafoundation/nova/runtime/util/FullAssetIdModel.kt
antonijzelinskij: Of course. We just need to add `kotlin-android-extensions` plugin

feature-account-api/src/main/java/io/novafoundation/nova/feature_account_api/data/model/Fee.kt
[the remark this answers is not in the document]
valentunn: Tbh this looks like half-way solution and will cause inconviniences for us when working with fees

feature-account-impl/src/main/java/io/novafoundation/nova/feature_account_impl/data/fee/utils/HydraDxQuoteSharedComputation.kt
```diff
@@ -35,31 +36,42 @@ class HydraDxQuoteSharedComputation(
...
         chain: Chain,
+        args: AssetExchangeQuoteArgs,
         accountId: AccountId,
-        fromAsset: Chain.Asset,
-        toAsset: Chain.Asset,
         scope: CoroutineScope
-    ): AssetExchangeQuote {
-        val key = "HydraDxQuote:${accountId.toHexString()}"
+    ): List<Path<HydraDxSwapEdge>> {
+        val key = "HydraDxPaths:${accountId.toHexString()}"
```
valentunn, line 45: The cache key should reflect all the arguments that might affect results
For example if you call the current code with different asset it will still returned old cached paths

feature-account-impl/src/main/java/io/novafoundation/nova/feature_account_impl/data/fee/utils/HydraDxQuoteSharedComputation.kt
```diff
@@ -35,31 +36,42 @@ class HydraDxQuoteSharedComputation(
...
             val assetConversion = getAssetConversion(chain, accountId, scope)
             val swapDirections = directions(chain, accountId, scope)
-            val paths = assetConversion.getPaths(swapDirections, args)
-            assetConversion.quote(paths, args)
+            assetConversion.getPaths(swapDirections, args)
         }
     }
 
-    private suspend fun getAssetConversion(
+    suspend fun quote(
```
valentunn, line 54: Why do we have two `quote` implementations and dublicated code of this implementation in ` HydraDxQuoteSharedComputation`

feature-account-api/src/main/java/io/novafoundation/nova/feature_account_api/data/model/Fee.kt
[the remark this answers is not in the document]
antonijzelinskij: I thought that it will be helpful to have separated interface but in UI pr I removed it since it's difficult to support.

Regarding using full chain.asset model in fee: We can create ChainAssetParcel (the same way like ChainParcel using serialization) and pass it to fee payloads. What do you think?

feature-account-impl/src/main/java/io/novafoundation/nova/feature_account_impl/data/fee/utils/HydraDxQuoteSharedComputation.kt
antonijzelinskij: Right. Thank you. I forgot to remove this one

valentunn (approved):

PR #1633, opened 2024-08-13
Feature/custom fee UI

antonijzelinskij:
#86958zzp8


stepanLav:
Task linked: [CU-86958zzp8 Android. Transfers. UI for Fees in custom asset](https://app.clickup.com/t/86958zzp8)

feature-wallet-api/src/main/java/io/novafoundation/nova/feature_wallet_api/presentation/mixin/fee/FeeLoaderMixin.kt
```diff
@@ -94,12 +115,10 @@ interface FeeLoaderMixin : GenericFeeLoaderMixin<SimpleFee> {
 
         fun loadFee(
             coroutineScope: CoroutineScope,
-            expectedChain: ChainId? = null,
```
antonijzelinskij, line 97: It seems that it wasn't used so I removed that

feature-staking-impl/src/main/java/io/novafoundation/nova/feature_staking_impl/presentation/staking/delegation/proxy/revoke/ConfirmRemoveStakingProxyViewModel.kt
```diff
@@ -133,7 +133,6 @@ class ConfirmRemoveStakingProxyViewModel(
 
             feeMixin.loadFee(
                 this,
-                chain.id,
```
antonijzelinskij, line 136: expected chain was used only here. But it's confirmation screen and chain here can not be changed.
It seems it was provided incorrectly

feature-account-api/src/main/java/io/novafoundation/nova/feature_account_api/data/model/Fee.kt
```diff
@@ -14,25 +18,31 @@ sealed interface Fee {
      * Information about origin that is supposed to send the transaction fee was calculated against
      */
     val submissionOrigin: SubmissionOrigin
+
+    val paymentAsset: PaymentAsset
```
valentunn, line 22: There is no point on distringuishing on data level between `Native` and `Asset` cases in a form of sealed class
`Fee`  should only be concerned about asset itself
Everything else is unnecessary complexity

feature-account-impl/src/main/java/io/novafoundation/nova/feature_account_impl/data/fee/utils/HydraDxQuoteSharedComputation.kt
```diff
@@ -79,9 +85,17 @@ class HydraDxQuoteSharedComputation(
         val key = "HydraDxAssetConversion:${accountId.toHexString()}"
 
         return computationalCache.useCache(key, scope) {
-            val storageSharedRequestBuilder = storageSharedRequestsBuilderFactory.create(chain.id)
+            val subscriptionBuilder = storageSharedRequestsBuilderFactory.create(chain.id)
             val assetConversion = assetConversionFactory.create(chain)
-            assetConversion.runSubscriptions(accountId, storageSharedRequestBuilder)
+
+            scope.launch {
```
valentunn, line 91: As I mentined in the previous PR review, you should use aggregate scope from the CompututationalCache and not the scope received in the arguments. Otherwise if the first scope that was used to initiate cache gets cancelled despite some other scopes still being present - subscriptions will be cancelled as well

feature-assets/src/main/java/io/novafoundation/nova/feature_assets/presentation/send/amount/SelectSendViewModel.kt
```diff
@@ -152,13 +154,19 @@ class SelectSendViewModel(
...
 
-    private val commissionAssetFlow = originChain.flatMapLatest(interactor::commissionAssetFlow)
-        .shareInBackground()
+    val originFeeMixin = feeLoaderMixinFactory.createGenericChangeableFee<OriginGenericFee>(
+        originAssetFlow,
+        coroutineScope,
+        configuration = Configuration(
+            initialState = Configuration.InitialState(
+                supportCustomFee = true
+            )
```
valentunn, line 166: I think better approach is to accept supportCustomFee: Flow<Boolean> instead of having a initial state and a setter. Its just easier to use

feature-assets/src/main/java/io/novafoundation/nova/feature_assets/presentation/send/amount/SelectSendViewModel.kt
```diff
@@ -544,3 +575,10 @@ class SelectSendViewModel(
         val balances: Asset?
     )
 }
+
+private class FeePayload(
```
valentunn, line 579: This is not used anywhere

feature-wallet-api/src/main/java/io/novafoundation/nova/feature_wallet_api/presentation/mixin/fee/FeeLoaderMixin.kt
```diff
@@ -76,7 +90,15 @@ interface GenericFeeLoaderMixin<F : GenericFee> : Retriable {
...
         suspend fun setFeeStatus(feeStatus: FeeStatus<F>)
 
-        fun invalidateFee()
+        suspend fun setSupportCustomFee(supportCustomFee: Boolean)
+
+        suspend fun invalidateFee()
+
+        suspend fun commissionChainAsset(): Chain.Asset
+
+        suspend fun commissionAsset(): Asset
```
valentunn, line 99: Those could be extensions based on `commissionAssetFlow`. We should keep interfaces as minimal as possible

feature-staking-impl/src/main/java/io/novafoundation/nova/feature_staking_impl/presentation/staking/delegation/proxy/revoke/ConfirmRemoveStakingProxyViewModel.kt
valentunn: It was also initially used on SelectSendViewModel untill we for some reason decided not to use loadFee extensions but regular combine
expected chain was introduced for cross chain transfers when origin chain can be changed (get DOT option on Swaps e.g.) to avoid race conditions related to fee loading & chain change

feature-wallet-api/src/main/java/io/novafoundation/nova/feature_wallet_api/presentation/mixin/fee/provider/ChangeableFeeLoaderProvider.kt
```diff
@@ -0,0 +1,293 @@
...
+import kotlinx.coroutines.flow.filterNotNull
+import kotlinx.coroutines.flow.first
+import kotlinx.coroutines.flow.flatMapLatest
+import kotlinx.coroutines.flow.launchIn
+import kotlinx.coroutines.flow.map
+import kotlinx.coroutines.flow.onEach
+import kotlinx.coroutines.launch
+import kotlinx.coroutines.withContext
+
+internal class ChangeableFeeLoaderProviderPresentation(
```
valentunn, line 43: There is no point on making a separate mixin. It is better to just modify original one, expecially given that you have modified the original interface
I wont review this file now as it has 90% LoC that has been copied from the original mixin

valentunn (changes requested):

feature-account-api/src/main/java/io/novafoundation/nova/feature_account_api/data/model/Fee.kt
antonijzelinskij: Okay. Fixed it.

feature-assets/src/main/java/io/novafoundation/nova/feature_assets/presentation/send/amount/SelectSendViewModel.kt
antonijzelinskij: It seems to me that we can leave it this way for now, because during the next scopes we will support custom fee everywhere, and this logic will be removed

feature-account-impl/src/main/java/io/novafoundation/nova/feature_account_impl/data/fee/utils/HydraDxQuoteSharedComputation.kt
```diff
@@ -64,13 +65,16 @@ class HydraDxQuoteSharedComputation(
             val subscriptionBuilder = storageSharedRequestsBuilderFactory.create(chain.id)
             val assetConversion = assetConversionFactory.create(chain)
 
-            scope.launch {
-                assetConversion.runSubscriptions(accountId, subscriptionBuilder)
-                    .throttleLast(500.milliseconds)
-                    .launchIn(scope)
+            // Required at least for stable swap
+            blockNumberUpdater.listenForUpdates(subscriptionBuilder, chain)
+                .launchIn(this)
```
antonijzelinskij, line 70: I think that for this case we can omit the update system

feature-account-impl/src/main/java/io/novafoundation/nova/feature_account_impl/data/fee/types/AssetConversionFeePayment.kt
```diff
@@ -81,21 +70,7 @@ internal class AssetConversionFeePayment(
...
-        val multiLocation = MultiLocation(
-            parents = BigInteger.ZERO,
-            interior = Junctions(
-                MultiLocation.Junction.PalletInstance(pallet.index),
-                MultiLocation.Junction.GeneralIndex(index)
-            )
-        )
-
-        return multiLocation.toEncodableInstance()
+        return multiLocationConverter.toMultiLocationOrThrow(paymentAsset).toEncodableInstance()
```
antonijzelinskij, line 73: To support such foreign assets like MYTH

valentunn: Good idea. We can go even futher and simplify ` encodableNativeAssetId` then by converting `chain.utilityAsset` with converter

feature-swap-impl/src/main/java/io/novafoundation/nova/feature_swap_impl/domain/swap/RealSwapService.kt
```diff
@@ -76,6 +76,12 @@ internal class RealSwapService(
         isCustomFeeToken && exchange.canPayFeeInNonUtilityToken(asset) && currentMetaAccount.type.requestedAccountPaysFees()
     }
 
+    override suspend fun sync(coroutineScope: CoroutineScope) {
+        exchanges(coroutineScope)
+            .values
+            .forEach { it.sync() }
```
valentunn, line 82: Lets make it async so some slow source wont stop everything else from syncing 
We have `mapAsync` helper for that

feature-swap-core/src/main/java/io/novafoundation/nova/feature_swap_core/data/assetExchange/conversion/types/hydra/RealHydraDxAssetConversion.kt
```diff
@@ -50,6 +50,10 @@ class RealHydraDxAssetConversion(
 
     private val conversionSources: List<HydraDxConversionSource> = createSources()
 
+    override suspend fun sync() {
+        conversionSources.forEach { it.sync() }
```
valentunn, line 54: Lets make it async so some slow source wont stop everything else from syncing
We have mapAsync helper for that

runtime/src/main/java/io/novafoundation/nova/runtime/network/updaters/BlockNumberUpdater.kt
```diff
@@ -1,24 +1,49 @@
...
-    storageCache: StorageCache
-) : SingleStorageKeyUpdater<Unit>(GlobalScope, crowdloanSharedState, chainRegistry, storageCache) {
+    private val chainRegistry: ChainRegistry,
+    private val storageCache: StorageCache
+) : Updater<Chain> {
 
-    override suspend fun storageKey(runtime: RuntimeSnapshot, scopeValue: Unit): String {
-        return runtime.metadata.system().storage("Number").storageKey()
-    }
+    override val scope: UpdateScope<Chain> = EmptyScope()
```
valentunn, line 26: EmptyScope? What is that? Seems like I missed something)
Is it renamed GlobalScope?

runtime/src/main/java/io/novafoundation/nova/runtime/network/updaters/BlockNumberUpdater.kt
```diff
@@ -1,24 +1,49 @@
...
+import kotlinx.coroutines.flow.onEach
 
 class BlockNumberUpdater(
-    chainRegistry: ChainRegistry,
-    crowdloanSharedState: SelectedAssetOptionSharedState<*>,
-    storageCache: StorageCache
-) : SingleStorageKeyUpdater<Unit>(GlobalScope, crowdloanSharedState, chainRegistry, storageCache) {
+    private val chainRegistry: ChainRegistry,
+    private val storageCache: StorageCache
+) : Updater<Chain> {
```
valentunn, line 24: Am I right that you removed SingleStorageKeyUpdater inheritance because it requires you to pass a chain id?
Thats actually deprecated since chain is now always available in `listenForUpdates`

Ideally we should remove it from SingleStorageKeyUpdater  but given that we have many updaters that inherts from it, there will be a lot of extra work to refact it now
So my suggestion is to make `chainIdHolder` filed optional in SingleStorageKeyUpdater   constructor and mark it as deprecated. This way, BlockNumberUpdater could just ignore it

valentunn (approved):

runtime/src/main/java/io/novafoundation/nova/runtime/network/updaters/BlockNumberUpdater.kt
antonijzelinskij: No this is cope that returns empty flow but typed.

I used it since scope here is unused now

runtime/src/main/java/io/novafoundation/nova/runtime/network/updaters/BlockNumberUpdater.kt
antonijzelinskij: Not really right. I just renamed old BlockNumberUpdater to SharedAssetBlockNumberUpdater and created BlockNumberUpdater that does not depend on the external state.

I think that this is not a very correct diff)

runtime/src/main/java/io/novafoundation/nova/runtime/network/updaters/BlockNumberUpdater.kt
valentunn: Is there a particular reason to not to use GlobalScope?

runtime/src/main/java/io/novafoundation/nova/runtime/network/updaters/BlockNumberUpdater.kt
valentunn: Okay I see

My suggestion seems to still be relevant though - instead of dublicating logic of  SingleStorageKeyUpdater  we could deprecate its "state" since chainIdHolder is no longer actually needed and use SingleStorageKeyUpdater   in a stateless way

runtime/src/main/java/io/novafoundation/nova/runtime/network/updaters/BlockNumberUpdater.kt
antonijzelinskij: Yes, Global scope is unusable here, since it's typed as Unit but I want to receive Chain in scopeValue in listenForUpdates

runtime/src/main/java/io/novafoundation/nova/runtime/network/updaters/BlockNumberUpdater.kt
antonijzelinskij: I think we can create a task and refactor it later. I think it's a minor issue

feature-account-api/src/main/java/io/novafoundation/nova/feature_account_api/data/fee/capability/CustomFeeAvailabilityFacade.kt
```diff
@@ -0,0 +1,13 @@
...
+
+interface CustomFeeCapability {
+
+    suspend fun canPayFeeInNonUtilityToken(chainAsset: Chain.Asset): Boolean
+}
+
+interface CustomFeeCapabilityFacade {
+
+    suspend fun canPayFeeInNonUtilityToken(chainAsset: Chain.Asset, customFeeCapability: CustomFeeCapability): Boolean
+}
```
valentunn, line 13: With current namings its a bit unclear what is the difference between them
I understand the difference itself from correpsponding implementations and overall idea to split per-chain specific logic and general one realted to proxies is good, its just naming that makes difference unclear

feature-wallet-api/src/main/java/io/novafoundation/nova/feature_wallet_api/presentation/mixin/fee/provider/ChangeableFeeLoaderProvider.kt
```diff
@@ -77,11 +77,10 @@ internal open class ChangeableFeeLoaderProvider<F : GenericFee>(
 
     private val supportCustomFeeFlow = MutableStateFlow(configuration.initialState.supportCustomFee)
 
-    private val availableFeeAssetsFlow: Flow<Map<Int, Chain.Asset>> = combine(supportCustomFeeFlow, selectedTokenFlow) { supportCustomFee, selectedToken ->
-        if (!supportCustomFee) return@combine emptyMap()
+    private val availableFeeAssetsFlow: Flow<Boolean> = combine(supportCustomFeeFlow, selectedTokenFlow) { supportCustomFee, selectedToken ->
```
valentunn, line 80: availableFeeAssetsFlow should be renamed if its now a Boolean

valentunn (approved):

feature-wallet-api/src/main/java/io/novafoundation/nova/feature_wallet_api/presentation/mixin/fee/provider/ChangeableFeeLoaderProvider.kt
antonijzelinskij: Agree. Fixed

feature-account-api/src/main/java/io/novafoundation/nova/feature_account_api/data/fee/capability/CustomFeeAvailabilityFacade.kt
antonijzelinskij: Yes, indeed, the name isn't clear enough, but it seems to me that everything becomes clear quite quickly if you open the implementation of the facade. I think we can leave it like this

PR #1641, opened 2024-08-22
bump: substrate sdk

stepanLav:
Screenshot 2024-08-22 at 10 20 34


valentunn:
Substrate SDK 2.1.3 seems to be corrupted a bit - there are seems to be missing features from 2.1.2. Thats why we rolled backup to 2.1.2 for now. Its on my TODO list to fix this

PR #1678, opened 2024-10-02
Track card creation state

antonijzelinskij:


common/src/main/java/io/novafoundation/nova/common/utils/formatting/duration/TimeDurationFormatter.kt
```diff
@@ -3,13 +3,33 @@ package io.novafoundation.nova.common.utils.formatting.duration
...
+    }
+
     override fun format(duration: Duration): String {
         return duration.toComponents { _, hours, minutes, seconds, _ ->
-            "%02d:%02d:%02d".format(hours, minutes, seconds)
+            val args = listOfNotNull(
+                hours.takeIf { useHours },
+                minutes.takeIf { useMinutes },
+                seconds.takeIf { useSeconds }
+            )
```
valentunn, line 24: This can be lazily cached to avoid this comutation on each `format`

feature-assets/src/main/java/io/novafoundation/nova/feature_assets/presentation/novacard/overview/NovaCardViewModel.kt
```diff
@@ -40,12 +42,25 @@ class NovaCardViewModel(
...
+        assetsRouter.openTopUpCard(payload)
+    }
+
+    fun onCardCreated() {
+        // No need to open/close timer dialog if card is already active
+        if (novaCardInteractor.isNovaCardStateActive()) return
+
+        novaCardInteractor.setNovaCardState(true)
+
+        closeTimerBottomSheet()
```
valentunn, line 60: Typically we dont do it so a screen closes another screen since it might lead to harder understanding of the flow
Given that Timer Bottom Sheet is a seaparate screen with its own view model, it could close itself upon noticing a nova card state being true

feature-assets/src/main/java/io/novafoundation/nova/feature_assets/presentation/novacard/overview/webViewController/WebViewCardCreationInterceptor.kt
```diff
@@ -0,0 +1,135 @@
...
+            .launchIn(coroutineScope)
+    }
+
+    /**
+     * @return true if the request was intercepted otherwise false
+     */
+    fun intercept(request: WebResourceRequest): Boolean {
+        val url = request.url.toString()
+
+        // Intercept requests that might return a JSON response
```
valentunn, line 64: We can remove those reduntant comments (I've pasted them together with other chatGPT code)

feature-assets/src/main/java/io/novafoundation/nova/feature_assets/presentation/novacard/overview/webViewController/WebViewCardCreationInterceptor.kt
```diff
@@ -0,0 +1,135 @@
...
+    private fun performOkHttpRequest(request: WebResourceRequest) {
+        try {
+            // Create OkHttp Request based on WebResourceRequest
+            val okHttpRequestBuilder = Request.Builder().url(request.url.toString())
+
+            // Set method (GET, POST, etc.) and request body if needed
+            when (request.method) {
+                "GET" -> okHttpRequestBuilder.get()
+                else -> okHttpRequestBuilder.get()
+            }
```
valentunn, line 82: Can be simplified to just `okHttpRequestBuilder.get()` since we dont expect anything else

feature-assets/src/main/java/io/novafoundation/nova/feature_assets/presentation/novacard/overview/webViewController/WebViewCardCreationInterceptor.kt
```diff
@@ -0,0 +1,135 @@
...
+        if (okHttpResponse.isSuccessful) {
+            val responseBody = okHttpResponse.body
+
+            val data = responseBody!!.byteStream().readText()
+            val cardsResponse = gson.fromJson<CardsResponse>(data)
+            val cards = cardsResponse.data
+            val containsMercurioCard = cards.any { it.issuedByMercurio }
+
+            if (containsMercurioCard) {
+                onCardCreatedListener.onCardCreated()
```
valentunn, line 120: This is an unexpected side effect, you wont expect `makeRequest` to notify the listener
Maybe we should separate `makeRequest` and listener notification? E.g. makeRequest already returns true, upper-level logic can use that to determine whether to invoke the listener or not, this method can be called something like `checkForCardCreation`

feature-assets/src/main/java/io/novafoundation/nova/feature_assets/presentation/novacard/waiting/WaitingNovaCardTopUpViewModel.kt
```diff
@@ -0,0 +1,43 @@
...
+        )
+
+        return EstimatedDurationFormatter(compoundFormatter)
+    }
+
+    fun timerFinished() {
+        showError(
+            resourceManager.getString(R.string.common_unexpected_error),
+            resourceManager.getString(R.string.fragment_waiting_top_up_time_out_error)
+        )
```
valentunn, line 39: I dont think it makes sense to show this scary message upon timer completion, lets clarify this behavior with Anton and Andrey

valentunn (approved):

feature-assets/src/main/java/io/novafoundation/nova/feature_assets/presentation/novacard/waiting/WaitingNovaCardTopUpViewModel.kt
```diff
@@ -0,0 +1,43 @@
...
+
+const val TIMER_MINUTES = 5
+
+class WaitingNovaCardTopUpViewModel(
+    private val assetsRouter: AssetsRouter,
+    private val resourceManager: ResourceManager
+) : BaseViewModel() {
+
+    fun getTimerValue(): TimerValue {
+        return TimerValue(TIMER_MINUTES.minutes.inWholeMilliseconds, System.currentTimeMillis())
```
valentunn, line 22: We shouldnt start a 5 minute timer each time we open the screen but rather use the value based on the time we set the "card is issuing" flag

common/src/main/res/values/strings.xml
```diff
@@ -1,6 +1,13 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <resources xmlns:tools="http://schemas.android.com/tools" tools:ignore="TypographyEllipsis">
 
+    <string name="waiting_top_up_card_timer">Estimated %s</string>
+    <string name="waiting_top_up_card_title">Your card being issued</string>
+    <string name="waiting_top_up_card_description">It can take up to 2 min. Please do not close this window while Mercuryo will receive your funds. This window will be closed automatically.</string>
```
valentunn, line 6: Should be 5 minutes

common/src/main/java/io/novafoundation/nova/common/utils/formatting/duration/TimeDurationFormatter.kt
antonijzelinskij: It's can't be cached since we can pass any duration to this function

PR #1681, opened 2024-10-04
Chore/refactor network managment

valentunn:
Fixes:

* NodeAutoBalance may crash the entire app if node iterator was empty
* Only one nodes is added to the local chain from preconfigured file

When fixing I noticed that is it very hard to keep whole network managment picture in mind due to broken responsibilities and invariants. Thus I refactored the following:

1. ChainRegistry no logner leaves db in inconsitent state (autoBalance=false, selectedNode=null). It now has a single point of entry to modify balancing settings via NodeSelectionStrategy
2.  Better types for Chain.Nodes: now we're explicit that AutoBalance is used only for ws nodes
3. Be explicit we are using unformatted NodeUrl everywhere it is needed
4. Refactored NodeSelectionSequenceStragey - firstly, the name was hacky, secondly, they no logner contain logic like "what should i do if no node was found". This logic has been moved to factory. Thirdly, it no longer needs satured nodes in `generate` as it is already available in the factory during creation
5. Some refactoring in ChainConnection to remove unnesseary triggers for node socket state changes
6. Chain.Nodes are no logner copied to filter for ws/http nodes. It is passed in the original form across the system. Only NodeSelectionStrategyProvider performs the fitlering


feature-settings-impl/src/main/java/io/novafoundation/nova/feature_settings_impl/domain/NetworkManagementChainInteractor.kt
```diff
@@ -77,11 +77,23 @@ class RealNetworkManagementChainInteractor(
 
     override suspend fun toggleAutoBalance(chainId: String) {
         val chain = chainRegistry.getChain(chainId)
-        chainRegistry.setAutoBalanceEnabled(chainId, chain.autoBalanceDisabled)
```
valentunn, line 80: `chain.autoBalanceDisabled` - this was extremely unclear, we should not use such sort of negation

runtime/src/main/java/io/novafoundation/nova/runtime/multiNetwork/connection/ChainConnection.kt
```diff
@@ -88,16 +86,21 @@ class ChainConnection internal constructor(
     ).shareIn(scope = this, started = SharingStarted.Eagerly)
 
     private val chain = MutableStateFlow(initialChain)
+
     private val availableNodes = chain.map { it.nodes }
-        .shareIn(scope = this, started = SharingStarted.Eagerly, replay = 1)
+        .distinctUntilChanged()
```
valentunn, line 91: We triggered node updates on each chain update

runtime/src/main/java/io/novafoundation/nova/runtime/multiNetwork/connection/ChainConnection.kt
```diff
@@ -88,16 +86,21 @@ class ChainConnection internal constructor(
...
+        .share(SharingStarted.Eagerly)
 
-    val currentUrl = chain.flatMapLatest { getNodeUrlFlow(it) }
-        .shareIn(scope = this, started = SharingStarted.Eagerly, replay = 1)
+    val currentUrl = nodeAutobalancer.connectionUrlFlow(
```
valentunn, line 94: We dont need to flatMap on chain since the `connectionUrlFlow` only uses chainId

runtime/src/main/java/io/novafoundation/nova/runtime/multiNetwork/connection/autobalance/NodeAutobalancer.kt
```diff
@@ -2,49 +2,41 @@ package io.novafoundation.nova.runtime.multiNetwork.connection.autobalance
...
 ) {
 
+    @OptIn(ExperimentalCoroutinesApi::class)
     fun connectionUrlFlow(
         chainId: ChainId,
         changeConnectionEventFlow: Flow<Unit>,
         availableNodesFlow: Flow<Chain.Nodes>,
     ): Flow<NodeWithSaturatedUrl?> {
-        return availableNodesFlow.flatMapLatest { nodesConfig ->
-            autobalanceStrategyProvider.strategyFlowFor(chainId, nodesConfig.nodeSelectionStrategy).transform { strategy ->
```
valentunn, line 29: No need to comlicate things with Flow here as we dont use it the way I originally expected

antonijzelinskij (approved):

PR #1682, opened 2024-10-04
Feature/swipe gov backend

antonijzelinskij:
#86961b694


feature-governance-impl/src/main/java/io/novafoundation/nova/feature_governance_impl/presentation/tindergov/cards/TinderGovCardsViewModel.kt
```diff
@@ -364,12 +291,38 @@ class TinderGovCardsViewModel(
...
+            ReferendaWithBasket(referendaCards, basket)
         }
             .addNewReferendaToCards()
             .launchIn(this)
     }
 
+    private fun List<ReferendumPreview>.mapToCardWithDetails(
+        summaries: Map<ReferendumId, String>,
+        amounts: Map<ReferendumId, AmountModel?>
+    ): List<CardWithDetails?> {
```
valentunn, line 310: Again, this abuses nullable values too much. This whole ViewModel is extremely hard to read because you cant quickly understand why certain values can be null and when they can be null

For example this method is called only once with later "filterNotNull" called afterwards. Why does it even return nullable values at the first place?

feature-governance-impl/src/main/java/io/novafoundation/nova/feature_governance_impl/presentation/tindergov/cards/model/CardWithDetails.kt
```diff
@@ -0,0 +1,25 @@
...
+            summary == other.summary &&
+            amount == other.amount
+    }
+
+    override fun hashCode(): Int {
+        var result = id.hashCode()
+        result = 31 * result + summary.hashCode()
+        result = 31 * result + amount.hashCode()
+        return result
+    }
```
valentunn, line 24: This really looks like a `data class`

valentunn (changes requested):

feature-governance-impl/src/main/java/io/novafoundation/nova/feature_governance_impl/presentation/tindergov/cards/TinderGovCardsViewModel.kt
```diff
@@ -102,12 +91,8 @@ class TinderGovCardsViewModel(
     private val _resetCards = MutableLiveData<Event<Unit>>()
     val resetCardsEvent: LiveData<Event<Unit>> = _resetCards
 
-    private val topReferendumWithDetails = combine(topCardIndex, cardsSummaryFlow, cardsAmountFlow) { topCardIndex, cardsSummary, cardsAmount ->
-        val referendum = sortedReferendaFlow.value.getOrNull(topCardIndex) ?: return@combine null
-        val cardSummary = cardsSummary.getOrDefault(referendum.id, ExtendedLoadingState.Loading)
-        val cardAmount = cardsAmount.getOrDefault(referendum.id, ExtendedLoadingState.Loading)
-
-        CardWithDetails(referendum, cardSummary, cardAmount)
+    private val topReferendumWithDetails = combine(topCardIndex, sortedReferendaFlow) { topCardIndex, referenda ->
```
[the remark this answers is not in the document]
antonijzelinskij: It's a top card of referendum. I renamed it to `topCardFlow` to make it more clear

valentunn (approved):

PR #1684, opened 2024-10-09
Fixed crash on delete chain

antonijzelinskij:
#8695mb1ky

### DIFF
--- a/feature-settings-impl/src/main/java/io/novafoundation/nova/feature_settings_impl/domain/NetworkManagementChainInteractor.kt
+++ b/feature-settings-impl/src/main/java/io/novafoundation/nova/feature_settings_impl/domain/NetworkManagementChainInteractor.kt
@@ -2,6 +2,7 @@ package io.novafoundation.nova.feature_settings_impl.domain
 
 import io.novafoundation.nova.common.utils.combine
 import io.novafoundation.nova.common.utils.flowOf
+import io.novafoundation.nova.feature_account_api.domain.interfaces.AccountInteractor
 import io.novafoundation.nova.runtime.ext.Geneses
 import io.novafoundation.nova.runtime.ext.genesisHash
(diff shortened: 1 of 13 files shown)


feature-settings-impl/src/main/java/io/novafoundation/nova/feature_settings_impl/domain/NetworkManagementChainInteractor.kt
```diff
@@ -62,7 +65,8 @@ interface NetworkManagementChainInteractor {
 class RealNetworkManagementChainInteractor(
     private val chainRegistry: ChainRegistry,
     private val nodeHealthStateTesterFactory: NodeHealthStateTesterFactory,
-    private val chainRepository: ChainRepository
+    private val chainRepository: ChainRepository,
+    private val accountInteractor: AccountInteractor
```
valentunn, line 69: In our current layers structure interactor should not depend on another interactor. Looks like it will be enough to just use `AccountRepository`

valentunn (approved):

feature-settings-impl/src/main/java/io/novafoundation/nova/feature_settings_impl/domain/NetworkManagementChainInteractor.kt
antonijzelinskij: Yes, agree, but I thought that it's not a repository responsibility to switch account if selected account  was deleted, so I used Interactor.

valentunn: Yeah indeed. But still we are moving to direction where interactor is only responsible for a certain screen (or flow)
What you are descriving sounds like a `UseCase` in our current terminology
Anyways  this is not so important

PR #1690, opened 2024-10-10
Block swipe gov for unsupported chains when proxied is using

antonijzelinskij:
#86964w7dt


feature-governance-impl/src/main/java/io/novafoundation/nova/feature_governance_impl/presentation/referenda/list/validation/StartSwipeGovValidationUi.kt
```diff
@@ -0,0 +1,22 @@
...
+fun handleStartSwipeGovValidationFailure(
+    resourceManager: ResourceManager,
+    validationStatus: ValidationStatus.NotValid<StartSwipeGovValidationFailure>
+): TransformedFailure {
+    return when (val reason = validationStatus.reason) {
+        is StartSwipeGovValidationFailure.NoChainAccountFound -> handleChainAccountNotFound(
+            failure = reason,
+            addAccountDescriptionRes = R.string.common_network_not_supported,
+            resourceManager = resourceManager,
+            goToWalletDetails = { }
```
valentunn, line 19: Hm ideally this argument should always trigger the navigation to wallet details to properly support "go to settings and add missing chain account" flow

valentunn (approved):

PR #1697, opened 2024-10-17
Portfolio View asset list

antonijzelinskij:
#86966j49v
#86966j49w


feature-assets/src/main/java/io/novafoundation/nova/feature_assets/domain/common/AssetPrefixNormalizing.kt
```diff
@@ -0,0 +1,26 @@
...
+import io.novafoundation.nova.common.utils.TokenSymbol
+import io.novafoundation.nova.common.utils.asTokenSymbol
+
+/**
+ * Some of tokens may have a prefix but may still be original token (like xcDOT, xcHDX)
+ * For some cases we need to normalize token symbol to original format
+ */
+
+private val normalizingPrefixList: List<String>
+    get() = listOf("xc")
```
valentunn, line 12: I dont think its good to have a list allocation each time we want to access it

feature-assets/src/main/java/io/novafoundation/nova/feature_assets/domain/common/AssetPrefixNormalizing.kt
```diff
@@ -0,0 +1,26 @@
...
+
+fun normalizeTokenSymbol(tokenSymbol: TokenSymbol): TokenSymbol {
+    return normalizeTokenSymbol(tokenSymbol.value).asTokenSymbol()
+}
```
valentunn, line 16: We can make it an extension function property for `TokenSymbol` to not to pollute "global namespace"

feature-assets/src/main/java/io/novafoundation/nova/feature_assets/domain/common/TokenAssetSorting.kt
```diff
@@ -0,0 +1,74 @@
...
+fun groupAndSortAssetsByToken(
+    assets: List<Asset>,
+    externalBalances: Map<FullChainAssetId, Balance>,
+    chainsById: Map<String, Chain>,
+    assetGroupComparator: Comparator<TokenAssetGroup> = getTokenAssetGroupBaseComparator(),
+    assetComparator: Comparator<AssetWithNetwork> = getTokenAssetBaseComparator()
+): Map<TokenAssetGroup, List<AssetWithNetwork>> {
+    return assets
+        .filter { chainsById.containsKey(it.token.configuration.chainId) }
+        .map { asset -> AssetWithNetwork(chainsById.getValue(asset.token.configuration.chainId), asset, asset.totalWithOffChain(externalBalances)) }
```
valentunn, line 42: This could be combined to `mapNotNull`

feature-assets/src/main/java/io/novafoundation/nova/feature_assets/domain/common/TotalAndTransferableBalance.kt
```diff
@@ -0,0 +1,33 @@
...
+
+    operator fun plus(other: TotalAndTransferableBalance): TotalAndTransferableBalance {
+        return TotalAndTransferableBalance(
+            total + other.total,
+            transferable + other.transferable
+        )
+    }
+}
+
+class Amount(
```
valentunn, line 22: I suggest to rename it to `PricedAmount`

feature-assets/src/main/java/io/novafoundation/nova/feature_assets/domain/common/TotalAndTransferableBalance.kt
```diff
@@ -0,0 +1,33 @@
+package io.novafoundation.nova.feature_assets.domain.common
+
+import java.math.BigDecimal
+
+class TotalAndTransferableBalance(
```
valentunn, line 5: `XAndY` kind of naming is usually hard to understand later on since it is long. Can we rename to something else?

feature-assets/src/main/java/io/novafoundation/nova/feature_assets/presentation/balance/common/AssetListMixin.kt
```diff
@@ -0,0 +1,115 @@
...
+
+    private val selectedCurrency = currencyInteractor.observeSelectCurrency()
+        .shareInBackground()
+
+    private val externalBalancesFlow = externalBalancesInteractor.observeExternalBalances()
+        .shareInBackground()
+
+    private val expandedTokenIdsFlow = MutableStateFlow(setOf<String>())
+
+    override val assetsViewModeFlow = assetsListInteractor.assetsViewModeFlow()
```
valentunn, line 76: Those two could be merged into

``` kotlin

sealed class AsetViewState {
  object Networks : AsetViewState ()
  
  class Tokens(val expandedTokens: Set<String>) 
}
```

To better reflect the fact that expanded tokens only make sense in a context of Tokens view

feature-assets/src/main/java/io/novafoundation/nova/feature_assets/presentation/balance/list/BalanceListViewModel.kt
```diff
@@ -89,9 +91,9 @@ class BalanceListViewModel(
         walletInteractor::syncAllNfts
     )
 
-    private val assetsFlow = walletInteractor.assetsFlow()
+    private val externalBalancesFlow = externalBalancesInteractor.observeExternalBalances()
```
valentunn, line 94: We also do that in `AssetListMixin`. Maybe worth passing this flow to the factory to avoid making two requests for the same data?

feature-assets/src/main/java/io/novafoundation/nova/feature_assets/presentation/balance/common/AssetGroupingDecoration.kt
```diff
@@ -111,13 +112,17 @@ class AssetGroupingDecoration(
...
+        return nextType == TYPE_NETWORK_GROUP || nextType == TYPE_TOKEN_GROUP || nextType == null
     }
 
     private fun shouldSkip(viewHolder: RecyclerView.ViewHolder): Boolean {
-        return viewHolder.bindingAdapterPosition == RecyclerView.NO_POSITION ||
-            (viewHolder !is AssetViewHolder && viewHolder !is AssetGroupViewHolder)
+        val isGroupViewHolder = viewHolder is NetworkAssetViewHolder ||
+            viewHolder is NetworkAssetGroupViewHolder ||
+            viewHolder is TokenAssetViewHolder ||
+            viewHolder is TokenAssetGroupViewHolder
```
valentunn, line 123: I suggest making a marker interface for all relevant ViewHolders like `AssetViewHolder` so that we dont need to modify this check in the future

feature-assets/src/main/java/io/novafoundation/nova/feature_assets/presentation/balance/common/AssetListMixin.kt
antonijzelinskij: I think it's not necessary here since it is using only in mixin that is simple enough

feature-assets/src/main/java/io/novafoundation/nova/feature_assets/presentation/balance/common/AssetGroupingDecoration.kt
antonijzelinskij: I fixed that in next prs

PR #1701, opened 2024-10-25
Add colored asset icon support

antonijzelinskij:
#86966j47d


feature-wallet-impl/src/main/java/io/novafoundation/nova/feature_wallet_impl/data/repository/RealChainAssetRepository.kt
```diff
@@ -26,6 +27,6 @@ class RealChainAssetRepository(
     }
 
     override suspend fun getEnabledAssets(): List<Chain.Asset> {
-        return chainAssetDao.getEnabledAssets().map { mapChainAssetLocalToAsset(it, gson) }
+        return chainAssetDao.getEnabledAssets().map { mapChainAssetLocalToAsset(it, gson, AssetIconMode.WHITE) }
```
antonijzelinskij, line 30: Oops, forgot to fix it.
Need to provide a service here

runtime/src/main/java/io/novafoundation/nova/runtime/multiNetwork/ChainRegistry.kt
```diff
@@ -63,11 +64,13 @@ class ChainRegistry(
...
-    private val gson: Gson
+    private val gson: Gson,
+    private val assetsIconModeService: AssetsIconModeService
 ) : CoroutineScope by CoroutineScope(Dispatchers.Default) {
 
-    val currentChains = chainDao.joinChainInfoFlow()
-        .mapList { mapChainLocalToChain(it, gson) }
+    val currentChains = combine(chainDao.joinChainInfoFlow(), assetsIconModeService.assetsIconModeFlow()) { chains, assetIconMode ->
+        chains.map { mapChainLocalToChain(it, gson, assetIconMode) }
+    }
```
antonijzelinskij, line 73: Provide the assetIconModeService here to useit in chain mapping

common/src/main/java/io/novafoundation/nova/common/data/model/AssetIconMode.kt
```diff
@@ -0,0 +1,5 @@
+package io.novafoundation.nova.common.data.model
+
+enum class AssetIconMode {
+    COLORED, WHITE
```
valentunn, line 4: Havent we been targeting to support any number of icon modes in the future?

common/src/main/java/io/novafoundation/nova/common/data/repository/AssetsIconModeService.kt
```diff
@@ -0,0 +1,54 @@
+package io.novafoundation.nova.common.data.repository
+
+import io.novafoundation.nova.common.data.model.AssetIconMode
+import io.novafoundation.nova.common.data.storage.Preferences
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.flow.Flow
+import kotlinx.coroutines.flow.map
+import kotlinx.coroutines.withContext
+
+interface AssetsIconModeService {
```
valentunn, line 10: This looks like a `Repository` and not a service

common/src/main/java/io/novafoundation/nova/common/data/repository/AssetsViewModeService.kt
```diff
@@ -7,17 +7,17 @@ import kotlinx.coroutines.flow.Flow
 import kotlinx.coroutines.flow.map
 import kotlinx.coroutines.withContext
 
-interface AssetsViewModeRepository {
+interface AssetsViewModeService {
```
valentunn, line 10: I think it should actually be `Repository`

runtime/src/main/java/io/novafoundation/nova/runtime/multiNetwork/chain/model/Chain.kt
```diff
@@ -46,6 +46,17 @@ data class Chain(
...
+    data class Icon(
+        val name: String,
+        val baseUrl: String,
+    ) {
+        companion object;
+
+        constructor(url: String) : this("", url)
+
+        fun getIconUrl(): String = "$baseUrl/$name"
+    }
```
valentunn, line 58: I understand that this is the easy  and quick way do achieve that. But it is not the right one.`ChainRegistry` and `Chain` already has too much responsibility and we definietely dont want to bring user prefererences in here.  We should have a separate entity, like `IconAppearanceProvider` that will construct icon urls based on chain and (internally) the mode and apply this change everywhere the icon is used. Thats what I meant under "it will require changes in a lot of places"

valentunn (changes requested):

PR #1702, opened 2024-10-28
Feature/receive screen update

antonijzelinskij:
#86966j486


common/src/main/java/io/novafoundation/nova/common/utils/images/Icon.kt
```diff
@@ -40,3 +40,9 @@ fun ImageView.setIconOrMakeGone(icon: Icon?, imageLoader: ImageLoader, builder:
 fun Drawable.asIcon() = Icon.FromDrawable(this)
 fun @receiver:DrawableRes Int.asIcon() = Icon.FromDrawableRes(this)
 fun String.asIcon() = Icon.FromLink(this)
+
+fun ImageLoader.Companion.formatIcon(icon: Icon): Any = when (icon) {
+    is Icon.FromDrawable -> icon.data
+    is Icon.FromDrawableRes -> icon.res
+    is Icon.FromLink -> icon.data
+}
```
valentunn, line 48: I would rather suggest to write a [mapper](https://coil-kt.github.io/coil/image_pipeline/#mappers) for Coil so we can directly pass `icon` instances to `load()`

valentunn (approved):

PR #1703, opened 2024-10-30
Feature/portfolio view search

antonijzelinskij:
#86966j48c


feature-assets/src/main/java/io/novafoundation/nova/feature_assets/presentation/balance/search/AssetSearchFragment.kt
```diff
@@ -81,15 +89,17 @@ class AssetSearchFragment :
...
         viewModel.searchResults.observe { data ->
             searchAssetsPlaceholder.setVisible(data.isEmpty())
             searchAssetList.setVisible(data.isNotEmpty())
 
-            assetsAdapter.submitListPreservingViewPoint(
-                data = data,
-                into = searchAssetList,
-                extraDiffCompletedCallback = { searchAssetList.invalidateItemDecorations() }
-            )
+            assetsAdapter.submitList(data) { searchAssetList.invalidateItemDecorations() }
```
antonijzelinskij, line 102: I don't use `submitListPreservingViewPoint` since it affects items animation.
Instead of that I reset recyclerview scroll position each time search query is updated

feature-assets/src/main/java/io/novafoundation/nova/feature_assets/presentation/balance/common/AssetExpandableAssetDecorationSettings.kt
```diff
@@ -0,0 +1,6 @@
+package io.novafoundation.nova.feature_assets.presentation.balance.common
+
+import android.view.animation.AccelerateDecelerateInterpolator
+import io.novafoundation.nova.common.utils.recyclerView.expandable.ExpandableAnimationSettings
+
+fun createAssetExpandableAnimationSettings() = ExpandableAnimationSettings(400, AccelerateDecelerateInterpolator())
```
valentunn, line 6: I suggest to put such "factory" functions as extensions to Companion object of the target create:

``` kotlin
ExpandableAnimationSettings.createForAssets()
```

This way we would not "pollute" the global namespace so much

feature-assets/src/main/java/io/novafoundation/nova/feature_assets/presentation/balance/common/ExpandableAssetsMixin.kt
```diff
@@ -0,0 +1,84 @@
...
+        return RealExpandableAssetsMixin(assetsFlow, currencyInteractor, assetsViewModeRepository)
+    }
+}
+
+interface ExpandableAssetsMixin {
+
+    val assetModelsFlow: Flow<List<BalanceListRvItem>>
+
+    fun expandToken(tokenGroupUi: TokenGroupUi)
+    suspend fun switchViewMode()
```
valentunn, line 32: Functions declarations should be separated by an empty line

valentunn (approved):

PR #1704, opened 2024-10-30
Feature/colored asset icons fix

antonijzelinskij:
#86966j47d


common/src/main/java/io/novafoundation/nova/common/di/CommonApi.kt
```diff
@@ -64,6 +66,36 @@ import java.util.Random
...
+    val automaticInteractionGate: AutomaticInteractionGate
+
+    val bannerVisibilityRepository: BannerVisibilityRepository
+
+    val provideActivityIntentProvider: ActivityIntentProvider
+
+    val googleApiAvailabilityProvider: GoogleApiAvailabilityProvider
+
+    val coinGeckoLinkParser: CoinGeckoLinkParser
+
```
valentunn, line 98: I dont mind this reordering but just of curiosity: what was the reason for doing that?

common/src/main/java/io/novafoundation/nova/common/presentation/AssetIconProvider.kt
```diff
@@ -0,0 +1,44 @@
...
+import io.novafoundation.nova.common.utils.images.Icon
+import io.novafoundation.nova.common.utils.images.asIcon
+
+interface AssetIconProvider {
+
+    val fallbackIcon: Icon
+
+    fun getAssetIcon(iconName: String?, fallback: Icon = fallbackIcon): Icon
+
+    fun getWhiteAssetIcon(iconName: String?, fallback: Icon = fallbackIcon): Icon
```
valentunn, line 15: I gave a few suggestions for the interface here:
1. The idea of having fallback icon being exposed by AssetIconProvider looks good. However I think we can make the the interface clearer if `getAssetIcon` wont allow for nullable `iconName`. Instead, we can create an extension function that allows nullability and has the same signature that is currently present in the interface directly. This way we will simplify the interface itself
2. Instead of creating a separate instance of `getWhiteAssetIcon` how about we make a function `getAssetIcon(iconName, iconType)` and implement`getWhiteAssetIcon` in a form of extension? This way you wont need to icon formatting logic twice since `getAssetIcon(iconName)` can easily delegate its work to `getAssetIcon(iconName, iconType)` by passing `iconType=selectedIconType`

feature-assets/src/main/java/io/novafoundation/nova/feature_assets/di/AssetsFeatureDependencies.kt
```diff
@@ -88,6 +89,88 @@ import javax.inject.Named
...
+    val chainStateRepository: ChainStateRepository
+
+    val holdsRepository: BalanceHoldsRepository
+
+    val holdsDao: HoldsDao
+
+    val coinGeckoLinkParser: CoinGeckoLinkParser
+
+    val assetIconProvider: AssetIconProvider
+
```
valentunn, line 173: Haha, those reorderings are the biggest part of PR diff :)

feature-assets/src/main/java/io/novafoundation/nova/feature_assets/domain/assets/list/AssetsListInteractor.kt
```diff
@@ -18,13 +18,13 @@ class AssetsListInteractor(
     private val accountRepository: AccountRepository,
     private val nftRepository: NftRepository,
     private val bannerVisibilityRepository: BannerVisibilityRepository,
-    private val assetsViewModeRepository: AssetsViewModeRepository
+    private val assetsViewModeService: AssetsViewModeRepository
```
valentunn, line 21: We should change it back to `assetsViewModeRepository` to much the class name

runtime/src/main/java/io/novafoundation/nova/runtime/multiNetwork/ChainRegistry.kt
```diff
@@ -66,8 +65,9 @@ class ChainRegistry(
     private val gson: Gson
 ) : CoroutineScope by CoroutineScope(Dispatchers.Default) {
 
-    val currentChains = chainDao.joinChainInfoFlow()
-        .mapList { mapChainLocalToChain(it, gson) }
+    val currentChains = chainDao.joinChainInfoFlow().map { chains ->
+        chains.map { mapChainLocalToChain(it, gson) }
+    }
```
valentunn, line 70: We can revert it back to `mapList`

valentunn (approved):

common/src/main/java/io/novafoundation/nova/common/di/CommonApi.kt
antonijzelinskij: I just refactored it to have fields above functions

common/src/main/java/io/novafoundation/nova/common/presentation/AssetIconProvider.kt
```diff
@@ -8,11 +8,11 @@ import io.novafoundation.nova.common.utils.images.asIcon
 
 interface AssetIconProvider {
 
-    val fallbackIcon: Icon
+    companion object;
 
-    fun getAssetIcon(iconName: String?, fallback: Icon = fallbackIcon): Icon
+    fun getAssetIconOrFallback(iconName: String): Icon
```
valentunn, line 13: Why there is `orFallback` in the name?

PR #1708, opened 2024-11-04
Feature/portfolio view UI improvements

antonijzelinskij:
#86966j497
#86966j49b


feature-assets/src/main/java/io/novafoundation/nova/feature_assets/presentation/balance/common/mappers/TokenAssetMappers.kt
```diff
@@ -18,23 +20,26 @@ import io.novafoundation.nova.feature_assets.presentation.balance.list.model.ite
 import io.novafoundation.nova.feature_assets.presentation.balance.list.model.items.TokenGroupUi
 import io.novafoundation.nova.feature_currency_api.presentation.formatters.formatAsCurrency
 import io.novafoundation.nova.feature_wallet_api.presentation.model.AmountModel
+import io.novafoundation.nova.feature_wallet_api.presentation.model.formatBalanceWithFraction
 
 fun GroupedList<TokenAssetGroup, AssetWithNetwork>.mapGroupedAssetsToUi(
```
valentunn, line 25: At this point I think we should create a separate `XXXFormatter` inteface + implementation for this to inject dependencies (resourceManager, assetIconProvider) instead of passing them to the mapper function

feature-wallet-api/src/main/java/io/novafoundation/nova/feature_wallet_api/presentation/model/AmountFormatters.kt
```diff
@@ -0,0 +1,31 @@
...
+import android.text.Spannable
+import android.text.SpannableStringBuilder
+import android.text.style.AbsoluteSizeSpan
+import android.text.style.ForegroundColorSpan
+import androidx.annotation.DimenRes
+import io.novafoundation.nova.common.resources.ResourceManager
+import io.novafoundation.nova.common.utils.formatting.toAmountWithFraction
+import io.novafoundation.nova.feature_wallet_api.R
+
+fun CharSequence.formatBalanceWithFraction(resourceManager: ResourceManager, @DimenRes floatAmountSize: Int): CharSequence {
```
valentunn, line 12: For the new formatters I suggest to always create a separate class so we dont need to pass dependencies as arguments. Using top-level formatters everywhere was a mistake and we are now experiencing the consequences with amount formatting for example

valentunn (approved):

PR #1716, opened 2024-11-10
Remove Multidex declaration

MGaetan89:
Since the min SDK is 24, it is no longer necessary to use the Multidex library.
The Robolectric-specific dependency is actually no longer used, but still declared in Gradle.

See the following for more info: https://developer.android.com/build/multidex#mdex-on-l


valentunn (approved):
Thanks!

PR #1729, opened 2024-11-25
Fix - handle evm addresses in polkadot namespace signing

valentunn:
Reproduced on https://mythical.market/


valentunn:
Closed in favour of #1730

PR #1733, opened 2024-11-27
Sdapp pr notes fix

antonijzelinskij:


common/src/main/java/io/novafoundation/nova/common/utils/CallbackLruCache.kt
```diff
@@ -0,0 +1,16 @@
+package io.novafoundation.nova.common.utils
+
+import android.util.LruCache
+
+class CallbackLruCache<K, V>(maxSize: Int) : LruCache<K, V>(maxSize) {
+
+    private var callback: ((V) -> Unit)? = null
```
valentunn, line 7: I think we can name it more specifically, `entryRemovedCallback`

common/src/main/java/io/novafoundation/nova/common/utils/InformationSize.kt
```diff
@@ -1,7 +1,7 @@
 package io.novafoundation.nova.common.utils
 
 @JvmInline
-value class InformationSize(private val sizeInBytes: Long) : Comparable<InformationSize> {
+value class InformationSize(val sizeInBytes: Long) : Comparable<InformationSize> {
```
valentunn, line 4: The idea behind this class is to mirror `Duration` functionality. So `sizeInBytes` should not be directly accessible but rather converted using e.g. `inWholeBytes` method (which is not present yet).
You can find more developed version of `InformationSize` [here](https://github.com/novasamatech/personhood-android/blob/main/common/src/main/java/com/polkadot/web3citizenship/common/utils/InformationSize.kt)

feature-dapp-impl/src/main/java/io/novafoundation/nova/feature_dapp_impl/data/repository/tabs/RealBrowserTabRepository.kt
```diff
@@ -20,6 +21,11 @@ class RealBrowserTabStorage(
         browserTabsDao.removeTab(tabId)
     }
 
+    override fun observeTabsWithNames(): Flow<Map<String, String?>> {
```
valentunn, line 24: Maybe `observeTabNames` or `observeTabNamesById` would be a better name?

feature-dapp-impl/src/main/java/io/novafoundation/nova/feature_dapp_impl/utils/tabs/TabMemoryRestrictionService.kt
```diff
@@ -2,14 +2,20 @@ package io.novafoundation.nova.feature_dapp_impl.utils.tabs
 
 import android.app.ActivityManager
 import android.content.Context
+import io.novafoundation.nova.common.utils.InformationSizeUnit
+import io.novafoundation.nova.common.utils.toInformationSize
 
 private const val MIN_TABS = 3
-private const val MEMORY_STEP = 100L * 1024L * 1024L // 100 Mb
+private val MEMORY_STEP = 100.toInformationSize(InformationSizeUnit.MEGABYTES).sizeInBytes
```
valentunn, line 9: `100.megabytes.inWholeBytes` reads much easier

feature-dapp-impl/src/main/java/io/novafoundation/nova/feature_dapp_impl/utils/tabs/models/BrowserTabSession.kt
```diff
@@ -4,67 +4,60 @@ import android.content.Context
...
     private var _webView: WebView? = null
-    val webView: WebView
-        get() {
-            return _webView ?: throw IllegalStateException("WebView is not initialized")
-        }
+    val webView: WebView by lazy {
+        initialize()
+        _webView!!.loadUrl(startUrl)
+        _webView!!
+    }
```
valentunn, line 42: It seems like the whole nullability and `initialize` workarounds are here just to initialize  `_webView` and `_webViewClient`  at the same time, when accessing either of them. Which is too implicit, like why would I expect one field to be initialized when accessing another? I suggest to do it in more explicit way:

``` kotlin
class SessionEnviroment(val webView: WebView, val client: Web3WebViewClient)

val enviroment: SessionEnviroment by lazy {
  createSessionEnviroment()
}
```

antonijzelinskij: Solve it in previous pr.
Since I pass webView via constructor it's not necessary to create a SessionEnviroment

common/src/main/java/io/novafoundation/nova/common/utils/CallbackLruCache.kt
```diff
@@ -4,13 +4,13 @@ import android.util.LruCache
 
 class CallbackLruCache<K, V>(maxSize: Int) : LruCache<K, V>(maxSize) {
 
-    private var callback: ((V) -> Unit)? = null
+    private var entryRevoewCallback: ((V) -> Unit)? = null
```
valentunn, line 7: `entryRemovedCallback`

valentunn (approved):

PR #1734, opened 2024-11-28
DApp open tabs animation

antonijzelinskij:
#8696nvafe

https://github.com/user-attachments/assets/ab3c5ad1-8335-4a8f-83e9-4a5467af62ba

### DIFF
--- a/feature-dapp-impl/src/main/java/io/novafoundation/nova/feature_dapp_impl/presentation/browser/main/DAppBrowserFragment.kt
+++ b/feature-dapp-impl/src/main/java/io/novafoundation/nova/feature_dapp_impl/presentation/browser/main/DAppBrowserFragment.kt
@@ -252,7 +252,6 @@ class DAppBrowserFragment : BaseFragment<DAppBrowserViewModel>(), OptionsBottomS
 
     private fun attachSession(session: BrowserTabSession) {
         clearProgress()
-        session.initialize(requireContext())
         session.attachToHost(createChromeClient(), this)
         webViewHolder.set(session.webView)
         webViewClient = session.webViewClient
--- a/feature-dapp-impl/src/main/java/io/novafoundation/nova/feature_dapp_impl/presentation/browser/main/DAppBrowserFragment.kt
+++ b/feature-dapp-impl/src/main/java/io/novafoundation/nova/feature_dapp_impl/presentation/browser/main/DAppBrowserFragment.kt
@@ -101,6 +101,7 @@ class DAppBrowserFragment : BaseFragment<DAppBrowserViewModel>(), OptionsBottomS
                         dappBrowserTransitionImage.animate()
                             .setDuration(300)
                             .alpha(0f)
+                            .withEndAction { dappBrowserTransitionImage.makeGone() }
                             .start()
                     }
                 )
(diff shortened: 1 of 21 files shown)


feature-dapp-impl/src/main/java/io/novafoundation/nova/feature_dapp_impl/presentation/browser/main/DAppBrowserFragment.kt
```diff
@@ -212,6 +251,7 @@ class DAppBrowserFragment : BaseFragment<DAppBrowserViewModel>(), OptionsBottomS
 
     private fun attachSession(session: BrowserTabSession) {
         clearProgress()
+        session.initialize(requireContext())
```
antonijzelinskij, line 254: We must initialize a session here to be sure that it happens in a thread with looper

valentunn: We can make factory method for `BrowserTabSession` to be suspend and switch to Main thread via `withContext` to ensure `BrowserTabSession` constructor runs in main thread

We should try to avoid late initialization as much as possible - delayed initialization makes it harder to reason about the class and introduces nothing but room for future bugs

feature-dapp-impl/src/main/java/io/novafoundation/nova/feature_dapp_impl/presentation/browser/main/DAppBrowserFragment.kt
```diff
@@ -112,6 +138,20 @@ class DAppBrowserFragment : BaseFragment<DAppBrowserViewModel>(), OptionsBottomS
...
+
+        setEnterSharedElementCallback(object : SharedElementCallback() {
+            override fun onSharedElementStart(
+                sharedElementNames: MutableList<String>?,
+                sharedElements: MutableList<View>?,
+                sharedElementSnapshots: MutableList<View>?
+            ) {
+                val sharedView = sharedElements?.firstOrNull { it.transitionName == DAPP_SHARED_ELEMENT_ID_IMAGE_TAB }
+                val sharedImageView = sharedView as? ImageView
+                dappBrowserTransitionImage.setImageDrawable(sharedImageView?.drawable) // Set image from shared element
```
valentunn, line 152: This is the solution to replace coil right? What was the issue with coil, can you elaborate?

feature-dapp-impl/src/main/java/io/novafoundation/nova/feature_dapp_impl/presentation/tab/BrowserTabsAdapter.kt
```diff
@@ -60,9 +59,7 @@ class BrowserTabViewHolder(
     fun bind(item: BrowserTabRvItem) = with(itemView) {
         browserTabCard.setOnClickListener { itemHandler.tabClicked(item, browserTabScreenshot) }
         browserTabClose.setOnClickListener { itemHandler.tabCloseClicked(item) }
-        browserTabScreenshot.load(item.tabScreenshotPath?.asFile(), imageLoader) {
-            transformations(TopCropTransformation())
```
valentunn, line 64: Can you elaborate why `TopCropTransform` cannot be used? Does it work bad during animation?

antonijzelinskij: Yes, the tab bounds and browser bounds may be different and it would be great to see how image expands in animation. It's not possible if we cut this image

valentunn (approved):

PR #1740, opened 2024-12-09
Fix/dapp fixes

antonijzelinskij:
#8696z16tb
#8696xky16
#8696xky56
- Remove sessions on account change
- Fixed back navigation for root nav holder
- Fixed navigation crashes
- Fixed initial page state providing to UI
- Fixed persistent state on page update
- Fixed search
- Sync dapps on search screen

### DIFF
--- a/feature-dapp-impl/src/main/java/io/novafoundation/nova/feature_dapp_impl/domain/RealBrowserSessionInteractor.kt
+++ b/feature-dapp-impl/src/main/java/io/novafoundation/nova/feature_dapp_impl/domain/RealBrowserSessionInteractor.kt
@@ -1,0 +1,22 @@
+package io.novafoundation.nova.feature_dapp_impl.domain
+
+import io.novafoundation.nova.common.utils.skipFirst
+import io.novafoundation.nova.feature_account_api.domain.interfaces.AccountRepository
+import io.novafoundation.nova.feature_dapp_api.domain.BrowserSessionInteractor
+import io.novafoundation.nova.feature_dapp_impl.utils.tabs.BrowserTabService
+import kotlinx.coroutines.flow.Flow
+import kotlinx.coroutines.flow.map
+import kotlinx.coroutines.flow.onEach
+
+class RealBrowserSessionInteractor(
+    private val browserTabService: BrowserTabService,
+    private val accountRepository: AccountRepository
+): BrowserSessionInteractor {
+
+    override fun destroyActiveSessionsOnAccountChange(): Flow<Unit> {
+        return accountRepository.selectedMetaAccountFlow()
+            .skipFirst() // Skip first to not trigger when app starts
+            .onEach { browserTabService.destroyActiveSessions() }
+            .map { Unit }
+    }
+}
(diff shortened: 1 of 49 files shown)


feature-dapp-impl/src/main/java/io/novafoundation/nova/feature_dapp_impl/utils/tabs/RealBrowserTabService.kt
```diff
@@ -45,8 +52,10 @@ class RealBrowserTabService(
 
     init {
         activeSessions.setCallback {
```
valentunn, line 54: I thinnk we should rename this to `setOnSessionRemoved` or smth similar. `setCallback` sounds too generic

feature-dapp-impl/src/main/java/io/novafoundation/nova/feature_dapp_impl/utils/tabs/RealBrowserTabService.kt
```diff
@@ -118,4 +129,27 @@ class RealBrowserTabService(
...
+            withContext(Dispatchers.Default) { browserTabInternalRepository.changeCurrentUrl(tabId, url) }
+            makeTabSnapshot(tabId)
+        }
+    }
+
+    private suspend fun makeTabSnapshot(tabId: String) {
+        val pageSession = activeSessions.get(tabId)
+
+        if (pageSession != null) {
+            val snapshot = pageSnapshotBuilder.getPageSnapshot(pageSession)
```
valentunn, line 151: Only this needs Main dispatcher right? I think we should just switch it here instead of switching multple times to Default

valentunn (approved):

feature-dapp-impl/src/main/java/io/novafoundation/nova/feature_dapp_impl/domain/RealBrowserSessionInteractor.kt
```diff
@@ -0,0 +1,22 @@
...
+import kotlinx.coroutines.flow.onEach
+
+class RealBrowserSessionInteractor(
+    private val browserTabService: BrowserTabService,
+    private val accountRepository: AccountRepository
+) : BrowserSessionInteractor {
+
+    override fun destroyActiveSessionsOnAccountChange(): Flow<Unit> {
+        return accountRepository.selectedMetaAccountFlow()
+            .skipFirst() // Skip first to not trigger when app starts
```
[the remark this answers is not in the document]
antonijzelinskij: Agree. Thank you

feature-dapp-impl/src/main/java/io/novafoundation/nova/feature_dapp_impl/utils/tabs/RealBrowserTabService.kt
antonijzelinskij: No, it's important to run it not in coroutine since sometimes we run it in pause and before onDestroyView. So If view destroyed during making a snapshot the snapshot may be drawn partly

PR #1746, opened 2025-01-02
Fix/dapp navigation

antonijzelinskij:
#8696rb6z7
#8697140h8
#86972enh0
#869714a36
#8697150vu

Since we now have 2 containers where we can open screens (ROOT and SPLIT_SCREEN), there is a need to contain 2 NavigationHolders, as well as launch the action for the correct NavController.

In addition, we want to make sure that when the action is launched, the new fragment will open in the currently attached navcontroller. SplitScreenNavController is detached every time we open a fragment in ROOT, so when we launch the action, it will automatically be executed for ROOT.

I moved this logic to NavigationBuilder. It is where the decision is made in which controller the fragment will be launched

### DIFF
--- a/app/src/main/java/io/novafoundation/nova/app/root/navigation/Ext.kt
+++ b/app/src/main/java/io/novafoundation/nova/app/root/navigation/Ext.kt
@@ -1,10 +1,16 @@
 package io.novafoundation.nova.app.root.navigation
 
 import android.annotation.SuppressLint
+import android.os.Bundle
 import androidx.annotation.IdRes
 import androidx.navigation.NavBackStackEntry
 import androidx.navigation.NavController
 import androidx.navigation.NavGraph
+import io.novafoundation.nova.app.R
+import io.novafoundation.nova.app.root.navigation.delayedNavigation.NavComponentDelayedNavigation
+import io.novafoundation.nova.app.root.navigation.navigators.BaseNavigator
+import io.novafoundation.nova.app.root.presentation.splitScreen.SplitScreenFragment
+import io.novafoundation.nova.app.root.presentation.splitScreen.SplitScreenPayload
 
 @SuppressLint("RestrictedApi")
 fun NavController.getBackStackEntryBefore(@IdRes id: Int): NavBackStackEntry {
@@ -21,4 +27,13 @@ fun NavController.getBackStackEntryBefore(@IdRes id: Int): NavBackStackEntry {
     }
 
     return backStack[previousIndex]
+}
+
+fun BaseNavigator.openSplitScreenWithInstantAction(actionId: Int, nestedActionExtras: Bundle? = null) {
+    val delayedNavigation = NavComponentDelayedNavigation(actionId, nestedActionExtras)
+
+    val splitScreenPayload = SplitScreenPayload.InstantNavigationOnAttach(delayedNavigation)
+    navigationBuilder(R.id.action_open_split_screen)
+        .setArgs(SplitScreenFragment.createPayload(splitScreenPayload))
+        .performInRoot()
 }
--- a/app/src/main/java/io/novafoundation/nova/app/root/navigation/Ext.kt
+++ b/app/src/main/java/io/novafoundation/nova/app/root/navigation/Ext.kt
@@ -35,5 +35,5 @@ fun BaseNavigator.openSplitScreenWithInstantAction(actionId: Int, nestedActionEx
     val splitScreenPayload = SplitScreenPayload.InstantNavigationOnAttach(delayedNavigation)
     navigationBuilder(R.id.action_open_split_screen)
         .setArgs(SplitScreenFragment.createPayload(splitScreenPayload))
-        .performInRoot()
+        .navigateInRoot()
 }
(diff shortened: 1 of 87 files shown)


app/src/main/java/io/novafoundation/nova/app/root/navigation/Ext.kt
```diff
@@ -22,3 +28,12 @@ fun NavController.getBackStackEntryBefore(@IdRes id: Int): NavBackStackEntry {
...
 }
+
+fun BaseNavigator.openSplitScreenWithInstantAction(actionId: Int, nestedActionExtras: Bundle? = null) {
+    val delayedNavigation = NavComponentDelayedNavigation(actionId, nestedActionExtras)
+
+    val splitScreenPayload = SplitScreenPayload.InstantNavigationOnAttach(delayedNavigation)
+    navigationBuilder(R.id.action_open_split_screen)
+        .setArgs(SplitScreenFragment.createPayload(splitScreenPayload))
+        .performInRoot()
+}
```
antonijzelinskij, line 39: Usefull to open main screen and instantly open another screen when nav controller will be attached

app/src/main/java/io/novafoundation/nova/app/root/presentation/splitScreen/SplitScreenViewModel.kt
```diff
@@ -72,4 +78,18 @@ class SplitScreenViewModel(
...
+                is SplitScreenPayload.InstantNavigationOnAttach -> {
+                    delayedNavigationRouter.runDelayedNavigation(it.delayedNavigation)
+                }
+
+                SplitScreenPayload.NoNavigation,
+                null -> {
+                } // Do nothing
+            }
+        }
+    }
```
antonijzelinskij, line 94: Use consumable payload yo instantly open another screen after nav controller is attached

app/src/main/java/io/novafoundation/nova/app/root/navigation/navigators/deepLinking/DeepLinkingNavigator.kt
```diff
@@ -1,36 +1,41 @@
...
     private val dAppRouter: DAppRouter,
-    private val governanceRouter: GovernanceRouter
-) : DeepLinkingRouter {
+    private val accountRouter: AccountRouter,
+    private val assetsRouter: AssetsRouter
+) : BaseNavigator(navigationHoldersRegistry), DeepLinkingRouter {
 
     override fun openAssetDetails(payload: AssetPayload) {
-        assetsRouter.openAssetDetails(payload)
+        openSplitScreenWithInstantAction(R.id.action_mainFragment_to_balanceDetailFragment, BalanceDetailFragment.getBundle(payload))
```
valentunn, line 26: Why do we use `withInstantAction` for an action that starts from `MainFragment` (action_mainFragment_to_balanceDetailFragment) already? Is there a typo in the action name and it is actually a global one?

app/src/main/java/io/novafoundation/nova/app/root/presentation/splitScreen/SplitScreenViewModel.kt
```diff
@@ -72,4 +78,18 @@ class SplitScreenViewModel(
             null
         )
     }
+
+    fun onNavigationAttached() {
+        consumablePayload.use {
```
valentunn, line 83: I suggest to rename it to `useOnce` and rename the `Consumer` class to `UseOnce` or something similar

app/src/main/java/io/novafoundation/nova/app/root/navigation/navigators/dApp/DAppNavigator.kt
```diff
@@ -14,71 +13,65 @@ import io.novafoundation.nova.feature_dapp_impl.presentation.search.DappSearchFr
...
 
     override fun openChangeAccount() {
-        mainNavigationHolder.navController?.performNavigation(R.id.action_open_switch_wallet)
+        navigationBuilder(R.id.action_open_switch_wallet)
+            .perform()
     }
 
     override fun openDAppBrowser(payload: DAppBrowserPayload, extras: FragmentNavigator.Extras?) {
-        // Close deapp browser if it is already opened
-        // TODO it's better to provide new url to existing browser
```
valentunn, line 27: The comment still looks relevant

common/src/main/java/io/novafoundation/nova/common/utils/Consumer.kt
```diff
@@ -0,0 +1,17 @@
+package io.novafoundation.nova.common.utils
+
+class Consumer<T>(private var target: T?) {
+
+    fun get(): T? {
+        val returnable = target
+        target = null
+        return returnable
```
valentunn, line 8:
```
return target.also {
target = null
}
```

common/src/main/java/io/novafoundation/nova/common/utils/Consumer.kt
```diff
@@ -0,0 +1,17 @@
...
+    }
+
+    fun use(block: (T) -> Unit) {
+        val target = get()
+        if (target != null) {
+            block(target)
+        }
```
valentunn, line 15: `get()?.let(block)`

valentunn (approved):
Looks solid overall, here are a few minor suggestions

antonijzelinskij:
Remove useAdd everywhere

antonijzelinskij:
rename main_nav_graph to split_screen_nav_graph

PR #1749, opened 2025-01-10
Feature/wallet associated browser

antonijzelinskij:
#8697cth21
#8697crak1


feature-dapp-impl/src/main/java/io/novafoundation/nova/feature_dapp_impl/utils/tabs/RealBrowserTabService.kt
```diff
@@ -125,16 +134,20 @@ class RealBrowserTabService(
...
+    override fun onPageChanged(tabId: String, url: String?, title: String?) {
+        if (url == null) return
+
+        launch {
+            withContext(Dispatchers.Default) {
+                activeSessions[tabId].currentUrl = url
+                browserTabInternalRepository.changeCurrentUrl(tabId, url)
+
+                val metadata = getKnownDappMetadata(url)
+                browserTabInternalRepository.changeKnownDAppMetadata(tabId, metadata?.iconLink)
```
valentunn, line 149: I think better approach would be to not to store (and update each time) the metadata for the dapp but rather get it from metadata repository directly when we need to use it. Otherwise we might encounter consistency issues since changeKnownDAppMetadata is only called when we open the relevant tab, but not, for example, when the new metadata is fetched. So we might face a situation when the tab list will show an old icon even after app restart and re-fetch of the new metadata

So e.g. interactor for tab list screen could use metadata repository to get all metadatas for currently present tabs and combine both pieces of data into a single model

valentunn (approved):

PR #1754, opened 2025-01-13
Fix/dapp navigation 2

antonijzelinskij:
I also refactored NavigationBuilder to avoid the possibility of creating a broken states.
So now we have 3 different navigation builders: for actions, cases, and graphs

https://app.clickup.com/t/86970whj2
https://app.clickup.com/t/8697a59um
https://app.clickup.com/t/8697a5cqj


app/src/main/java/io/novafoundation/nova/app/root/navigation/navigators/dApp/DAppNavigator.kt
```diff
@@ -39,41 +38,72 @@ class DAppNavigator(
...
+            NavOptions.Builder()
+                .setEnterAnim(R.anim.fragment_slide_in)
+                .setExitAnim(R.anim.fragment_open_exit)
+                .setPopEnterAnim(R.anim.fragment_close_enter)
+                .setPopExitAnim(R.anim.fragment_slide_out)
+                .setPopUpTo(R.id.splitScreenFragment, false)
+        }
+
+        return setNavOptions(navOptionsBuilder.build())
+    }
```
antonijzelinskij, line 108: I found a lot of bugs before in dapp navigation that is catched because navigation component doesn't support circled navigation.

The solution is using direct program navigation to graphs. So here I setup a nav animations

app/src/main/java/io/novafoundation/nova/app/root/navigation/navigators/builder/GraphNavigationBuilder.kt
```diff
@@ -0,0 +1,14 @@
+package io.novafoundation.nova.app.root.navigation.navigators.builder
+
+import io.novafoundation.nova.app.root.navigation.holders.NavigationHolder
+import io.novafoundation.nova.app.root.navigation.navigators.NavigationHoldersRegistry
+
+class GraphNavigationBuilder(
```
valentunn, line 6: I suggest to rename it to `OpenGraphNavigationBuilder`

app/src/main/java/io/novafoundation/nova/app/root/navigation/navigators/dApp/DAppNavigator.kt
```diff
@@ -39,41 +38,72 @@ class DAppNavigator(
...
+                .setPopExitAnim(R.anim.fragment_slide_out)
+                .setPopUpTo(R.id.splitScreenFragment, false)
+        } else {
+            // Slide in/out animations
+            NavOptions.Builder()
+                .setEnterAnim(R.anim.fragment_slide_in)
+                .setExitAnim(R.anim.fragment_open_exit)
+                .setPopEnterAnim(R.anim.fragment_close_enter)
+                .setPopExitAnim(R.anim.fragment_slide_out)
+                .setPopUpTo(R.id.splitScreenFragment, false)
```
valentunn, line 104: Why do we always pop up to splitScreenFragment?

valentunn (approved):

app/src/main/java/io/novafoundation/nova/app/root/navigation/navigators/dApp/DAppNavigator.kt
antonijzelinskij: To close dapp browser when use back navigation

valentunn: So this is basically to pop any dapp browser-related screens? Should we pop to the dapp browser graph insttead?
Also I think we should rename this method which is currently named "setDappAnimations" but it not only sets the animation, but alters the navigation itself

PR #1758, opened 2025-01-14
Refactor low-level xcm logic

valentunn:
* Move everything related to bare xcm to a separate feature module for better reusability
* Get rid of separate VersionedXXX types, introduce a generic `VersionedXcm<T>` type, which has a companion interface `VersionedToDynamicScaleInstance` that allows to encode different `T` types in a context of a specific xcm version
* Split `MultiLocation` to `RelativeMultiLocation` and `AbsoluteMultiLocation`. `RelativeMultiLocation` is a full analogue of what `MultiLocation` was before. `AbsoluteMultiLocation` is a new type that explicitly denotes that this location is absolute, meaning it is presented from the "root point", so it cannot have parents. This allows to implement general-purpose `reanchor` method easily.
* Get rid of deprecated `MultiAsset.Id` and `VersionedMultiAssetId`. Replace them with simpler `MultiAssetId` and `VersionedXcm<MultiAssetId>`


feature-wallet-api/build.gradle
```diff
@@ -31,6 +31,7 @@ dependencies {
     implementation materialDep
 
     implementation daggerDep
+    implementation project(':feature-xcm:api')
```
antonijzelinskij, line 34: Let's move it to other project dependencies?

antonijzelinskij (approved):

feature-wallet-api/build.gradle
valentunn: Yeah thats how Android Studio performs automatic dependency addition :/

PR #1761, opened 2025-01-15
Remove tabs only for current meta account

antonijzelinskij:


core-db/src/main/java/io/novafoundation/nova/core_db/dao/BrowserTabsDao.kt
```diff
@@ -19,8 +22,8 @@ abstract class BrowserTabsDao {
     @Query("DELETE FROM browser_tabs WHERE id = :tabId")
     abstract suspend fun removeTab(tabId: String)
 
-    @Query("DELETE FROM browser_tabs")
-    abstract suspend fun removeAllTabs()
+    @Query("DELETE FROM browser_tabs WHERE id IN (:tabIds)")
+    abstract suspend fun removeTabsByIds(tabIds: List<String>)
```
valentunn, line 26: Can be simplified to `DELETE FROM browser_tabs WHHERE metaId = :metaId`

app/src/main/java/io/novafoundation/nova/app/root/domain/SplitScreenInteractor.kt
```diff
@@ -17,6 +17,7 @@ class SplitScreenInteractor(
     }
 
     suspend fun removeAllTabs() {
-        repository.removeAllTabs()
+        val metaAccount = accountRepository.getSelectedMetaAccount()
+        repository.removeTabsForMetaAccount(metaAccount.id)
```
valentunn, line 21: Why do we sometimes use tabs repository and sometimes tab service? I think the usages should be consistent

core-db/src/main/java/io/novafoundation/nova/core_db/dao/BrowserTabsDao.kt
antonijzelinskij: I know but I use two separated methods to understand which tabs should close active sessions

valentunn: Delete function can return ids of deleted entries

antonijzelinskij: As I know it works only for @Delete annotation or for Room 3.35+ with RETURNING keyword.

We can increase version but I think current solution is pretty well

app/src/main/java/io/novafoundation/nova/app/root/domain/SplitScreenInteractor.kt
antonijzelinskij: It was too difficult to move BrowserTabService to dapp-api module since a lot of dependencies should be moved too.

antonijzelinskij: I moved all logic for remove into repository so now it's not duplicated.
I also think it's pretty well solution

PR #1762, opened 2025-01-15
Feature/max button 

antonijzelinskij:
#8696nafzv


feature-governance-impl/src/main/java/io/novafoundation/nova/feature_governance_impl/presentation/delegation/delegation/create/chooseAmount/NewDelegationChooseAmountViewModel.kt
```diff
@@ -131,10 +151,10 @@ class NewDelegationChooseAmountViewModel(
         originFeeMixin.connectWith(
             inputSource1 = amountChooserMixin.backPressuredAmount,
             inputSource2 = selectedConvictionFlow,
-            scope = this,
-            feeConstructor = { amount, conviction ->
+            inputSource3 = chainFlow,
+            feeConstructor = { feePaymentCurrency, amount, conviction, chain ->
                 interactor.estimateFee(
-                    amount = amount.toPlanks(),
+                    amount = feePaymentCurrency.toFeePaymentAsset(chain).planksFromAmount(amount),
```
valentunn, line 157: I feel like using `FeePaymentCurrency = Native | Custon(Asset)` was not a great idea since now I can see we cannot really extract `asset` from it without passing `chain` from somewhere. I think at some point we should switch to inline class that wraps `Chain.Asset`. Wdyt?

stepanLav:
Task linked: [CU-8696nafzv Android. Max button. Append it to all action screen](https://app.clickup.com/t/8696nafzv)

stepanLav:
Task linked: [CU-86976ju7r iOS. Search. No daps categories](https://app.clickup.com/t/86976ju7r)

feature-governance-impl/src/main/java/io/novafoundation/nova/feature_governance_impl/presentation/delegation/delegation/create/chooseAmount/NewDelegationChooseAmountViewModel.kt
valentunn: Mentioned issue is solved by introducing `amountInputMixin.backpressuredPlanks` since we should not rely on the fee mixin data at the first place for planks conversion

PR #1771, opened 2025-01-20
Myth staking/setup start staking

valentunn:
* Based on StartParachainStakingFragment, create a generic `StartSingleSelectStakingViewModel<T>` that implements the main presenation logic for "single stake target selection". Note that the a particular case (e.g. Parachain, Mythos) logic is extracted into a separate, `StartSingleSelectStakingLogic` interface instead of declaring abract methods in  `StartSingleSelectStakingViewModel`. This is done to avoid dealing with not yet properly built ViewModel object - during `init` blocks of superclass fields e.t.c. of subclasses are not yet initialized
* Introduce the same generalization for `SingleSelectRecommendator` (from `CollatorRecommendator`) and `SingleSelectStakingRewardEstimationComponentFactory` (from `ParachainStakingRewardsComponentFactory`)
* Implement most of the remaining data sources and repositories, including collator fetching, user's stake distribution
* `StartSingleSelectStakingViewModel` is not yet complete as it needs proper collator selection screen and confirm screen. This is to come in the next PRs
image


feature-staking-impl/src/main/java/io/novafoundation/nova/feature_staking_impl/domain/rewards/RewardCalculator.kt
```diff
@@ -5,6 +5,8 @@ import java.math.BigDecimal
 class PeriodReturns(
     val gainAmount: BigDecimal,
     val gainFraction: BigDecimal,
+    // true = APY, false = APR
+    val isCompound: Boolean,
```
antonijzelinskij, line 9: Maybe it's better to have enum value here?
And this field will be like that: `val isCompound = type == Type.APY`

antonijzelinskij (approved):

feature-staking-impl/src/main/java/io/novafoundation/nova/feature_staking_impl/domain/rewards/RewardCalculator.kt
valentunn: APR vs APY and compound: true vs false are completely equal terms, I've explained in the comment just for the references for potentially unaware reader. This sort of explanation is still required even if we convert it to enum

PR #1772, opened 2025-01-20
Hotfix/dapp browser fix

antonijzelinskij:


valentunn (approved):

feature-dapp-impl/src/main/java/io/novafoundation/nova/feature_dapp_impl/data/repository/RealDAppMetadataRepository.kt
```diff
@@ -24,6 +24,10 @@ class RealDAppMetadataRepository(
 
     private val dappMetadatasFlow = MutableSharedFlow<DappCatalog>(replay = 1, onBufferOverflow = BufferOverflow.DROP_OLDEST)
 
+    override suspend fun isDAppsSynced(): Boolean {
+        return dappMetadatasFlow.first().dApps.isNotEmpty()
```
valentunn, line 28: This will hang indefinitely if dapps are not synced since `dappMetadataFlow` is shared flow and not a state flow

antonijzelinskij: Thank you. Change it to replyCache

feature-dapp-impl/src/main/java/io/novafoundation/nova/feature_dapp_impl/utils/tabs/RealBrowserTabService.kt
```diff
@@ -82,12 +81,14 @@ class RealBrowserTabService(
             metaId = accountRepository.getSelectedMetaAccount().id,
             pageSnapshot = PageSnapshot.fromName(Urls.domainOf(url)),
             currentUrl = url,
-            knownDAppMetadata = getKnownDappMetadata(url),
+            knownDAppMetadata = null,
```
valentunn, line 84: I don't think we should do this just to call `syncKnownDAppMetadata`. This approach lacks readability as it splits the tab initialization into two steps

antonijzelinskij:
Release severity: Normal
Release notes:
- Fixed opening DApps in more staking options

PR #1784, opened 2025-01-27
Myth staking/unbond

valentunn:


common/src/main/res/color/input_monocrhome_icon_tint.xml
```diff
@@ -0,0 +1,5 @@
+<?xml version="1.0" encoding="utf-8"?>
+<selector xmlns:android="http://schemas.android.com/apk/res/android">
+    <item android:color="#7AFFFFFF" android:state_enabled="false" />
+    <item android:color="#FFF" />
```
antonijzelinskij, line 4: We have icon_primary and icon_pressed colors in res. Let's use it please to avoid manual refactoring in future

feature-wallet-api/src/main/res/layout/view_choose_amount_input.xml
```diff
@@ -19,7 +19,8 @@
         app:layout_constraintBottom_toBottomOf="parent"
         app:layout_constraintStart_toStartOf="parent"
         app:layout_constraintTop_toTopOf="parent"
-        tools:src="@drawable/ic_token_ksm" />
+        tools:src="@drawable/ic_token_ksm"
+        app:tint="@color/input_monocrhome_icon_tint" />
```
antonijzelinskij, line 23: We may use colored icons for this field. I think we will get some problems with tinting here.

Are you sure we need to use tint here?

feature-staking-impl/src/main/java/io/novafoundation/nova/feature_staking_impl/presentation/mythos/unbond/setup/SetupUnbondMythosViewModel.kt
```diff
@@ -0,0 +1,223 @@
...
+    private val selectedCollatorIdFlow = selectedCollatorFlow.map { it.accountId }
+
+    private val selectedCollatorWithStake = combine(selectedCollatorFlow, currentDelegatorStateFlow) { selectedCollator, currentDelegatorState ->
+        val stake = currentDelegatorState.delegationAmountTo(selectedCollator.accountId)
+
+        TargetWithStakedAmount(stake, selectedCollator)
+    }
+
+    private val stakedAmount = selectedCollatorWithStake.map { it.stake }
+
```
antonijzelinskij, line 95: Should we add shareInBackground for both flows?
We use them in few places

feature-wallet-api/src/main/res/layout/view_choose_amount_input.xml
antonijzelinskij: As I see we can remove tinting here and remove this color res since we already use alpha for this image when it's disabled

valentunn: Let me re-check this, I might have changed the approach mid-way

PR #1787, opened 2025-01-28
Redeem flow

valentunn:


app/src/main/java/io/novafoundation/nova/app/root/navigation/navigators/staking/mythos/MythosStakingNavigator.kt
```diff
@@ -53,7 +56,17 @@ class MythosStakingNavigator(
...
+            .action(R.id.action_stakingFragment_to_mythosRedeemFragment)
+            .navigateInFirstAttachedContext()
+    }
+
+    override fun finishRedeemFlow(redeemConsequences: RedeemConsequences) {
+        if (redeemConsequences.willKillStash) {
+            stakingDashboardRouter.returnToStakingDashboard()
+        } else {
+            returnToStakingMain()
+        }
```
antonijzelinskij, line 69: I think it would be better to provide a boolean flag here instead of domain model

app/src/main/res/navigation/staking_main_graph.xml
```diff
@@ -531,4 +539,9 @@
     <include app:graph="@navigation/staking_mythos_start_graph" />
 
     <include app:graph="@navigation/staking_mythos_unbond_graph" />
+
+    <fragment
+        android:id="@+id/mythosRedeemFragment"
+        android:name="io.novafoundation.nova.feature_staking_impl.presentation.mythos.redeem.MythosRedeemFragment"
+        android:label="MythosRedeemFragment" />
 </navigation>
```
antonijzelinskij, line 547: Let's organize a navigation graph to not mix includes, actions and fragments.

I suggest this approach: includes first, than actions and then fragments

antonijzelinskij (approved):

app/src/main/java/io/novafoundation/nova/app/root/navigation/navigators/staking/mythos/MythosStakingNavigator.kt
valentunn: Yeah the typelias to boolean seems to be a better fit, but it is not a new model and was used already in similar redeem flows so I just reused that

PR #1800, opened 2025-02-12
Feature/banners

antonijzelinskij:
#8697mdwy5


stepanLav:
Task linked: [CU-8697mdwy5 Android. Banners. DApp/Browser and Assets tab](https://app.clickup.com/t/8697mdwy5)

feature-banners-impl/src/main/java/io/novafoundation/nova/feature_banners_impl/presentation/banner/PromotionBannersMixin.kt
```diff
@@ -0,0 +1,108 @@
...
+    private val imageLoader: ImageLoader,
+    private val context: Context,
+    private val bannersSource: BannersSource
+) : PromotionBannersMixin {
+
+    override val bannersFlow = bannersSource.observeBanners()
+        .map { banners ->
+            val resources = loadResources(banners)
+            banners.map { mapBanner(it, resources) }
+        }.withSafeLoading()
```
valentunn, line 57: I suggest we always do `share/shareInBackground` in a mixin. We can do that by passing `CoroutineContext` to the mixin factory and then to the mixin itself so mixin is a fully-featured component similar to ViewModel that does not require any extra actions on top of just creating it and wiring with UI

feature-banners-impl/src/main/java/io/novafoundation/nova/feature_banners_impl/presentation/banner/PromotionBannersMixin.kt
```diff
@@ -0,0 +1,108 @@
...
+        promotionBannersInteractor.closeBanner(banner.id)
+    }
+
+    override fun handleBannerAction(page: BannerPageModel) {
+        val url = page.actionUrl ?: return
+
+        try {
+            val intent = Intent(Intent.ACTION_VIEW, Uri.parse(url))
+                .addFlags(FLAG_ACTIVITY_NEW_TASK)
+            context.startActivity(intent)
```
valentunn, line 69: Feels like this could be an extension, smth like `launchDeeplink`?

feature-banners-impl/src/main/java/io/novafoundation/nova/feature_banners_impl/presentation/banner/PromotionBannersMixin.kt
```diff
@@ -0,0 +1,108 @@
...
+        } catch (e: Exception) {
+            Log.e(LOG_TAG, "Error while running an activity", e)
+        }
```
valentunn, line 72: Feels like this could be an extension, smth like `launchDeeplink`?

feature-banners-impl/src/main/java/io/novafoundation/nova/feature_banners_impl/presentation/banner/PromotionBannersMixin.kt
```diff
@@ -0,0 +1,108 @@
...
+            addAll(banners.map { it.backgroundUrl })
+        }
+
+        val loadingImagesResult = imagesSet.toMap {
+            val imageRequest = ImageRequest.Builder(context)
+                .data(it)
+                .build()
+
+            asyncWithContext { imageLoader.execute(imageRequest) }
+        }
```
valentunn, line 87: Are we sure we cant show banners lazily? This approach seems to be quite slow and unreliable (failure of loading one banner will cause the entire loading to fail)

valentunn (approved):
Good job overall!

feature-banners-impl/src/main/java/io/novafoundation/nova/feature_banners_impl/presentation/banner/PromotionBannersMixin.kt
antonijzelinskij: Yes we discussed it and decide to show it when all data is loaded

PR #1801, opened 2025-02-14
Fixed tab session creation

antonijzelinskij:


valentunn (approved):

feature-dapp-impl/src/main/java/io/novafoundation/nova/feature_dapp_impl/utils/tabs/RealBrowserTabService.kt
```diff
@@ -132,12 +136,14 @@ class RealBrowserTabService(
...
-        return CurrentTabState.Selected(
-            tab,
-            activeSessions[tabId] ?: addNewSession(tab)
-        )
+        return createTabMutex.withLock {
+            val tabId = selectedTabId ?: return CurrentTabState.NotSelected
+            val tab = allTabs[tabId] ?: return CurrentTabState.NotSelected
+            CurrentTabState.Selected(
+                tab,
+                activeSessions[tabId] ?: addNewSession(tab)
```
valentunn, line 144: I think we should just put this line under mutex since the rest is thread-safe

feature-dapp-impl/src/main/java/io/novafoundation/nova/feature_dapp_impl/utils/tabs/RealBrowserTabService.kt
```diff
@@ -53,7 +58,7 @@ class RealBrowserTabService(
             tabs = allTabs.values.toList(),
             selectedTab = currentTabState(selectedTabId, allTabs)
         )
-    }
+    }.share()
```
valentunn, line 61: `shareInBackground`, why would we want to execute that on main thread?

PR #1814, opened 2025-02-21
Fixed bottom sheet on asset details

antonijzelinskij:


feature-assets/src/main/java/io/novafoundation/nova/feature_assets/presentation/transaction/history/TransactionHistoryView.kt
```diff
@@ -75,17 +81,29 @@ class TransferHistorySheet @JvmOverloads constructor(
...
         View.inflate(context, R.layout.view_transfer_history, this)
 
-        background = context.getTopRoundedCornerDrawable(fillColorRes = R.color.secondary_screen_background, cornerSizeInDp = 16)
+        val contentBackgroundDrawable = context.getTopRoundedCornerDrawable(
+            fillColorRes = R.color.android_history_background,
+            strokeColorRes = R.color.container_border,
+            cornerSizeInDp = 16
+        )
+
+        // Extend background drawable from left and right to make stroke in background on sides hidden
```
valentunn, line 100: Should we rather request Andrey to provide background without the stroke to avoid this workaround?

valentunn (approved):

feature-assets/src/main/java/io/novafoundation/nova/feature_assets/presentation/transaction/history/TransactionHistoryView.kt
antonijzelinskij: Maybe we may just create an extension named `withStrokeOutside` that may be useful in some cases

valentunn: I think it's important to try to keep things simpler. 
The very idea of correcting not ideal assets in the code instead of requesting fixed assets looks like a redundant overhead to me

antonijzelinskij: As we discussed with designer it's not good solution to remove stroke here, since is some cases this bs may overlap other widgets and content may not look not very good

PR #1835, opened 2025-03-03
Feature/ledger connection modes

antonijzelinskij:
#8697krwtm


stepanLav:
Task linked: [CU-8697krwtm Android. Ledger. Remove Bluetooth requirement and fetch proper name by USB](https://app.clickup.com/t/8697krwtm)

feature-ledger-api/src/main/java/io/novafoundation/nova/feature_ledger_api/sdk/discovery/DiscoveryMethod.kt
```diff
@@ -0,0 +1,11 @@
+package io.novafoundation.nova.feature_ledger_api.sdk.discovery
+
+enum class DiscoveryMethod {
+    BLE,
+    USB,
+    ALL
```
valentunn, line 6: I think all should be replaced with `List<DiscoveryMethod>`

feature-ledger-impl/src/main/java/io/novafoundation/nova/feature_ledger_impl/presentation/account/common/selectLedger/SelectLedgerViewModel.kt
```diff
@@ -190,7 +210,11 @@ abstract class SelectLedgerViewModel(
             listOf(Manifest.permission.ACCESS_FINE_LOCATION)
         }
 
-        val granted = permissionsAsker.requirePermissionsOrExit(*permissions.toTypedArray())
+        val granted = if (discoveryMethod.isBluetoothRequired()) {
+            permissionsAsker.requirePermissionsOrExit(*permissions.toTypedArray())
+        } else {
+            permissionsAsker.requirePermissionsOrIgnore(*permissions.toTypedArray())
+        }
```
valentunn, line 217: I think we should not hardcode that only bluetooth cannot work without required permission

In general, we want to avoid `isBluetooth` at all. This should rather be done by implementing **general** behavior like 

``` kotlin
class DiscoveryMethodPermissions(val required: Boolean, val permissions: List<String>)

fun DiscoveryMethid.requiredPermissions(): DiscoveryMethodPermissions
```

then we can do
``` kotlin
val permissions = discoverMethods.map { it.requiredPermissions() }
val systemPermissions = permissions.flatMap { it.permissions }.distinct()
val shouldExitUponReject = permissions.any { it.required } // this could be better implemented to not to exit when only non-required permissions are rejected but I showed the full analogue of PR's logic
```

Please make sure all the code in this PR follows this idea

feature-ledger-impl/src/main/java/io/novafoundation/nova/feature_ledger_impl/presentation/account/common/selectLedger/stateMachine/states/SelectDiscoveryModeState.kt
```diff
@@ -0,0 +1,18 @@
...
+import io.novafoundation.nova.feature_ledger_impl.presentation.account.common.selectLedger.stateMachine.SelectLedgerEvent
+import io.novafoundation.nova.feature_ledger_impl.presentation.account.common.selectLedger.stateMachine.SideEffect
+
+class SelectDiscoveryModeState : SelectLedgerState() {
+
+    override suspend fun StateMachine.Transition<SelectLedgerState, SideEffect>.performTransition(event: SelectLedgerEvent) {
+        if (event is SelectLedgerEvent.DiscoveryMethodSelected) {
+            when (event.discoveryMethod) {
+                DiscoveryMethod.BLE, DiscoveryMethod.ALL -> WaitingForPermissionsState(event.discoveryMethod)
+                DiscoveryMethod.USB -> startDiscovery(event.discoveryMethod)
```
valentunn, line 14: Again, we should not hardcode that a particular discovery method requires permissions and others - don't

feature-ledger-impl/src/main/java/io/novafoundation/nova/feature_ledger_impl/presentation/account/common/selectLedger/stateMachine/states/SelectLedgerState.kt
```diff
@@ -11,32 +14,32 @@ import java.io.InvalidObjectException
...
-            DiscoveryRequirement.BLUETOOTH in missingRequirements -> SideEffect.EnableBluetooth
-            DiscoveryRequirement.LOCATION in missingRequirements -> SideEffect.EnableLocation
-            else -> {
-                val requirementsJoinedToString = missingRequirements.joinToString { it.toString() }
-                throw InvalidObjectException("missingRequirements contains values that aren't processing: $requirementsJoinedToString")
+        if (discoveryMethod.isBluetoothUsing()) {
+            val sideEffect = getSideEffectFromRequirements(missingRequirements)
+
+            if (discoveryMethod.isBluetoothRequired()) {
+                emitState(MissingDiscoveryRequirementState(missingRequirements, discoveryMethod))
```
valentunn, line 38: No hardcode of bluetooth please. This is not extendable

valentunn (changes requested):

valentunn (approved):

PR #1847, opened 2025-03-07
Changed banners specification

antonijzelinskij:


common/src/main/java/io/novafoundation/nova/common/utils/recyclerView/space/SpaceBetween.kt
```diff
@@ -0,0 +1,37 @@
...
+import android.view.View
+import androidx.recyclerview.widget.RecyclerView
+import io.novafoundation.nova.common.utils.dp
+import io.novafoundation.nova.common.utils.recyclerView.ViewType
+
+class SpaceBetween(
+    private val upperViewType: ViewType,
+    private val lowerViewType: ViewType,
+    private val spaceDp: Int
+) : RecyclerViewItemSpace {
```
antonijzelinskij, line 13: Maybe you will suggest a better name, since actually it's "space between" only when upper item on top and lower item is going next so if their positions are different - the spacing will not be applied

feature-assets/src/main/java/io/novafoundation/nova/feature_assets/presentation/balance/list/BalanceListFragment.kt
```diff
@@ -222,6 +224,24 @@ class BalanceListFragment :
         viewModel.swapClicked()
     }
 
+    private fun setupRecyclerViewSpacing() {
+        balanceListAssets.addSpaceItemDecoration {
+            addSpaceBetween(AssetsHeaderAdapter.getViewType(), PromotionBannerAdapter.getViewType(), spaceDp = 4)
+            addSpaceBetween(PromotionBannerAdapter.getViewType(), ManageAssetsAdapter.getViewType(), spaceDp = 4)
+            addSpaceBetween(AssetsHeaderAdapter.getViewType(), ManageAssetsAdapter.getViewType(), spaceDp = 24)
+        }
```
antonijzelinskij, line 232: I tried to set up spacing comparing view holder types (when get them from findViewHolderFromLayoutPosition) but it isn't works well so doing the same but with view types is much stable

feature-assets/src/main/java/io/novafoundation/nova/feature_assets/presentation/balance/list/view/AssetsHeaderAdapter.kt
```diff
@@ -16,7 +17,13 @@ import kotlinx.android.synthetic.main.item_asset_header.view.balanceListTotalBal
 import kotlinx.android.synthetic.main.item_asset_header.view.balanceListTotalTitle
 import kotlinx.android.synthetic.main.item_asset_header.view.balanceListWalletConnect
 
-class AssetsHeaderAdapter(private val handler: Handler) : RecyclerView.Adapter<HeaderHolder>() {
+class AssetsHeaderAdapter(private val handler: Handler) : RecyclerView.Adapter<AssetsHeaderHolder>() {
+
+    companion object {
+        const val VIEW_TYPE = 100
```
antonijzelinskij, line 23: Unfortunately we need to check manually to keep view types different in different Adapters

common/src/main/java/io/novafoundation/nova/common/utils/recyclerView/space/SpaceBetween.kt
valentunn: I think the name is good

common/src/main/java/io/novafoundation/nova/common/utils/recyclerView/space/SpaceBetween.kt
```diff
@@ -0,0 +1,37 @@
...
+
+        return false
+    }
+
+    private fun shouldSetSpaceForItems(
+        parent: RecyclerView,
+        view: View
+    ): Boolean {
+        parent.getViewType(view).takeIf { it == upperViewType } ?: return false
+        parent.getNextViewType(view).takeIf { it == lowerViewType } ?: return false
```
valentunn, line 29: I think its better to explicitly include non-null check here - it took me two attempts to understand we use `takeIf` because `getViewType` is nullable
So

`currentViewType != null && currentViewType == upperViewType`

common/src/main/java/io/novafoundation/nova/common/utils/recyclerView/space/SpaceItemDecoration.kt
```diff
@@ -0,0 +1,47 @@
...
+
+import android.graphics.Rect
+import android.view.View
+import androidx.recyclerview.widget.RecyclerView
+import androidx.recyclerview.widget.RecyclerView.ItemDecoration
+import io.novafoundation.nova.common.utils.recyclerView.ViewType
+
+class SpaceItemDecoration(private val spaces: List<RecyclerViewItemSpace>) : ItemDecoration() {
+
+    class Builder {
```
valentunn, line 11: While this approach reflects the current design requirements well, I feel like the overall idea of "offset depends on a pair of items" might become quite complex and hard to handle once the number of view types. Maybe we should work with design team to decide on "context-independent" margins instead? So a margin for a certain item doesn't depend on surrounding items

feature-assets/src/main/java/io/novafoundation/nova/feature_assets/presentation/balance/list/view/AssetsHeaderAdapter.kt
valentunn: A best practice is to ui layout id of the item instead. Those are guaranteed to be stable and unique so we wont ever face any dublicates e.t.c.

feature-assets/src/main/java/io/novafoundation/nova/feature_assets/presentation/balance/list/BalanceListFragment.kt
```diff
@@ -222,6 +224,24 @@ class BalanceListFragment :
         viewModel.swapClicked()
     }
 
+    private fun setupRecyclerViewSpacing() {
+        balanceListAssets.addSpaceItemDecoration {
+            addSpaceBetween(AssetsHeaderAdapter.getViewType(), PromotionBannerAdapter.getViewType(), spaceDp = 4)
```
valentunn, line 229: I think this can be a little bit clearer if we create some interface

``` kotlin
interface WithViewType {
  val viewType: Int
}
```

and let each relevant adapter's companion object implement it

This way we can then do

``` kotlin
addSpaceBetween(AssetsHeaderAdapter, PromotionBannerAdapter, spaceDp = 4)
```

valentunn (approved):

feature-assets/src/main/java/io/novafoundation/nova/feature_assets/presentation/balance/list/view/AssetsHeaderAdapter.kt
antonijzelinskij: Yes, thank you

common/src/main/java/io/novafoundation/nova/common/utils/recyclerView/space/SpaceItemDecoration.kt
antonijzelinskij: I think this idea is good, but if we provide for this, let's specify that this is desirable but optional since we have the ability to have different spacing between elements to may leave the designers the option to use it if necessary

feature-assets/src/main/java/io/novafoundation/nova/feature_assets/presentation/balance/list/BalanceListFragment.kt
antonijzelinskij: I think that idea is nice, I'll provide that!

valentunn (approved):

PR #1848, opened 2025-03-07
Fix/charts long click

antonijzelinskij:


common/src/main/AndroidManifest.xml
```diff
@@ -1 +1,4 @@
-<manifest />
+<manifest xmlns:android="http://schemas.android.com/apk/res/android">
+
+    <uses-permission android:name="android.permission.VIBRATE" />
```
valentunn, line 3: We usually gather all permissions in the App module to quickly be able to check all existing ones

feature-assets/src/main/java/io/novafoundation/nova/feature_assets/presentation/views/priceCharts/ChartController.kt
```diff
@@ -86,21 +91,41 @@ class ChartController(private val chart: LineChart, private val callback: Callba
...
-                }
+        chart.setOnTouchListener(this)
+    }
+
+    override fun onTouch(v: View, event: MotionEvent): Boolean {
+        longClickDetector.onTouch(v, event)
+
+        when (event.action) {
+            MotionEvent.ACTION_UP, MotionEvent.ACTION_CANCEL -> {
+                isLongClickDetected = false
```
valentunn, line 102: I think this should be detelegated to `longClickDetector` which can inform about aborted long click via another callback. Otherwise we are separating detection logic between `longClickDetector` and this class

feature-assets/src/main/java/io/novafoundation/nova/feature_assets/presentation/views/priceCharts/ChartController.kt
```diff
@@ -86,21 +91,41 @@ class ChartController(private val chart: LineChart, private val callback: Callba
...
+
+        return true
+    }
 
-            false
+    fun onChartsLongClick(event: MotionEvent) {
+        isLongClickDetected = true
+        handleChartsTouch(event)
+
+        context.vibrate(50) // Add very short vibration on long click
```
valentunn, line 116: What is the reason for not using our `DeviceVibrator` here?

valentunn (approved):

common/src/main/AndroidManifest.xml
antonijzelinskij: Yes, I know. I did it so that the error wouldn't be highlighted in the common module, because it seems that in this case the permissions need to be written in the current module, even if they are in app

feature-assets/src/main/java/io/novafoundation/nova/feature_assets/presentation/views/priceCharts/ChartController.kt
antonijzelinskij: I think that LongClickDetector sholdn't contains any state. The state is important only for this controller, since after long click we may handle any other motion events

feature-assets/src/main/java/io/novafoundation/nova/feature_assets/presentation/views/priceCharts/ChartController.kt
antonijzelinskij: I just don't like to inject something inside views

common/src/main/AndroidManifest.xml
valentunn: I think thats just Android Studio not being able to comperehend multi-module setup proprely

feature-assets/src/main/java/io/novafoundation/nova/feature_assets/presentation/views/priceCharts/ChartController.kt
antonijzelinskij: Hmm, Although LongPressDetector already has a state.
Okay it's valid. I'll provide this changes

feature-assets/src/main/java/io/novafoundation/nova/feature_assets/presentation/views/priceCharts/ChartController.kt
valentunn: I don't think thats the issue. On the other hand, using different approaches to achieve one thing in different places is the issue as it creates a mess on a larger scale

feature-assets/src/main/java/io/novafoundation/nova/feature_assets/presentation/views/priceCharts/ChartController.kt
valentunn: 1. It already contains the state, so I didn't quite get the argument
2. I am not saying it should contain the state, just that it is strange to seprate long click detection between two classes. If the entity is named "LongClickDetector" it should do everything reletaed to long click detection (both start and end) right?

feature-assets/src/main/java/io/novafoundation/nova/feature_assets/presentation/views/priceCharts/ChartController.kt
antonijzelinskij: Yes, I agree, that's why I created a shared extension method and use it in both implementations. 
I think that in this case such an implementation is acceptable. What do you think?

PR #1849, opened 2025-03-07
Fix swap quote loading on a fresh install

valentunn:
**Core issue:** `BlockNumberUpdater` was running only on origin chain, preventing `StableSwapSource` to construct its internal data structures since it was waiting for block number to arrive. This might result in quoting to be completely blocked in case block number for Hydratin was never synced (e.g. if you open DOT Polkadot -> PAH on a fresh start)

**Fix:** completely get rid of necessity of running BlockTimeUpdater by subscribing directly to the remote storage. Of course we need to thing about reusing such sort of common subscribtions and requesting them in a lazy way. This is implemented by introducing `SharedSwapSubscriptions` abstraction and `LazyAsyncCache` that is used in its implementation

**Future work:** I also wanted to get rid of SwapUpdateSystem entirely to simplify the model even further but it is harder to do since `SwapUpdateSystem` syncs `system.account` which is used in `canBalanceDropBelowEd` which is used in multiple places, including validations

#86981r3fn

### DIFF
--- a/common/src/main/java/io/novafoundation/nova/common/data/memory/LazyAsyncCache.kt
+++ b/common/src/main/java/io/novafoundation/nova/common/data/memory/LazyAsyncCache.kt
@@ -1,0 +1,72 @@
+package io.novafoundation.nova.common.data.memory
+
+import io.novafoundation.nova.common.utils.invokeOnCompletion
+import kotlinx.coroutines.CoroutineScope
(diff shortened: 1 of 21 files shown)


valentunn:
#86981r3fn

stepanLav:
Task linked: [CU-86981r3fn Android. Constant loading state on swap PAH to Polkadot](https://app.clickup.com/t/86981r3fn)

PR #1864, opened 2025-03-14
Base nova card

antonijzelinskij:


common/src/main/java/io/novafoundation/nova/common/utils/systemCall/SystemCallExecutor.kt
```diff
@@ -5,40 +5,82 @@ import io.novafoundation.nova.common.resources.ContextManager
...
+        class Suspendable<T>(
+            private val continuation: Continuation<Result<T>>,
+            override val systemCall: SystemCall<T>
+        ) : PendingRequest<T> {
+            override fun onResult(result: Result<T>) {
+                continuation.resume(result)
+            }
+        }
+
+        class Normal<T>(
```
antonijzelinskij, line 29: Maybe you will suggest a better name for this

valentunn: Looks like we ont need a `sealed` class here, we can just store `() -> Unit` callback which `suspend` will fill as `{ continuation.resume(result)` and `Normal` will just use regular callback

Regarding the name, I would use `Callback` if we need so

feature-assets/src/main/java/io/novafoundation/nova/feature_assets/data/repository/NovaCardStateRepository.kt
```diff
@@ -0,0 +1,64 @@
...
+        return NovaCardState.valueOf(novaCardState)
+    }
+
+    override fun setNovaCardCreationState(state: NovaCardState) {
+        preferences.putString(PREFS_NOVA_CARD_STATE, state.toString())
+    }
+
+    override fun observeNovaCardCreationState(): Flow<NovaCardState> {
+        return preferences.keyFlow(PREFS_NOVA_CARD_STATE)
+            .map { getNovaCardCreationState() }
```
valentunn, line 46: Seems like this could be `stringFlow().map {}` instead

feature-assets/src/main/java/io/novafoundation/nova/feature_assets/domain/novaCard/NovaCardInteractor.kt
```diff
@@ -0,0 +1,74 @@
...
+
+import io.novafoundation.nova.feature_assets.data.repository.NovaCardStateRepository
+import kotlinx.coroutines.flow.Flow
+import kotlin.time.Duration.Companion.minutes
+
+enum class NovaCardState {
+    NONE,
+    CREATION,
+    CREATED
+}
```
valentunn, line 11: Better to extract to a separate file

feature-assets/src/main/java/io/novafoundation/nova/feature_assets/domain/novaCard/NovaCardInteractor.kt
```diff
@@ -0,0 +1,74 @@
...
+    override fun observeNovaCardState(): Flow<NovaCardState> {
+        return novaCardStateRepository.observeNovaCardCreationState()
+    }
+
+    override fun setLastTopUpTime(time: Long) {
+        novaCardStateRepository.setLastTopUpTime(time)
+    }
+
+    override fun getLastTopUpTime(): Long {
+        val cardBeingIssuedTime = novaCardStateRepository.getLastTopUpTime()
```
valentunn, line 67: This looks strange. While do we put `novaCardStateRepository.getLastTopUpTime()` intao a variable named `cardBeingIssuedTime`?

feature-assets/src/main/java/io/novafoundation/nova/feature_assets/domain/novaCard/NovaCardInteractor.kt
```diff
@@ -0,0 +1,74 @@
...
+
+    override fun observeNovaCardState(): Flow<NovaCardState> {
+        return novaCardStateRepository.observeNovaCardCreationState()
+    }
+
+    override fun setLastTopUpTime(time: Long) {
+        novaCardStateRepository.setLastTopUpTime(time)
+    }
+
+    override fun getLastTopUpTime(): Long {
```
valentunn, line 66: I dont get this logic overall:

1. We get the last recorded top up time from repository, add a timeout value it, so it is a `cardCreationTimetout` lets say
2. We substract current time from it, so it is `timeUntilCreationTimeout`

Why is it named `getLastTopUpTime`?

feature-assets/src/main/java/io/novafoundation/nova/feature_assets/domain/send/SendInteractor.kt
```diff
@@ -43,14 +43,18 @@ class SendInteractor(
...
-            val submissionFee = getAssetTransfers(transfer).calculateFee(transfer, coroutineScope = coroutineScope)
             TransferFee(
-                originFee = OriginFee(submissionFee, null),
+                originFee = getOriginFee(transfer, coroutineScope),
                 crossChainFee = null
             )
         }
     }
 
+    suspend fun getOriginFee(transfer: AssetTransfer, coroutineScope: CoroutineScope): OriginFee = withContext(Dispatchers.Default) {
```
valentunn, line 53: I dont think we should use `OriginFee` since it also contains informaition about cross chain transfers which is irrelevant for the regular transfer. We can just use `SubmissionFee` returned from `etAssetTransfers(transfer).calculateFee`

feature-assets/src/main/java/io/novafoundation/nova/feature_assets/presentation/novacard/overview/NovaCardViewModel.kt
```diff
@@ -0,0 +1,112 @@
...
+
+    override fun onCardCreated() {
+        setCardStateCreated()
+    }
+
+    private suspend fun getTopUpChain(): Chain {
+        return chainRegistry.getChain(ChainGeneses.POLKADOT)
+    }
+
+    private fun checkCardCreationState() {
```
valentunn, line 97: I think the name of the function should indicate that it might open a bottom sheet
Something like `ensureCardCreationIsBlocking`?

feature-assets/src/main/java/io/novafoundation/nova/feature_assets/presentation/novacard/overview/webViewController/NovaCardWebViewController.kt
```diff
@@ -0,0 +1,65 @@
...
+) {
+
+    fun create(
+        interceptors: List<NovaCardInterceptor>,
+        setupConfig: CardSetupConfig,
+        scope: CoroutineScope,
+    ): NovaCardWebViewController {
+        return NovaCardWebViewController(
+            novaCardWebViewClient = novaCardWebViewClientFactory.create(interceptors),
+            novaCardWebChromeClient = novaCardWebChromeClientFactory.create(scope),
```
valentunn, line 24: So both of those factories seems to only be used here, so maybe we can reduce the number of abstractions to simplify the understanding by embedding those two factories logic here, in the `NovaCardWebViewControllerFactory`?

But the overall idea of moving creation of something out of ViewModel is nice. Its just I think here we dont need to move the "inner" logic to separate entities

feature-assets/src/main/java/io/novafoundation/nova/feature_assets/presentation/novacard/overview/webViewController/interceptors/CardCreationInterceptor.kt
```diff
@@ -0,0 +1,67 @@
...
+) : NovaCardInterceptor {
+
+    interface Callback {
+
+        fun onCardCreated()
+    }
+
+    /**
+     * @return true if the request was intercepted otherwise false
+     */
```
valentunn, line 30: Lets revisit the comments in this file. They seem to be redundant

feature-assets/src/main/java/io/novafoundation/nova/feature_assets/presentation/novacard/overview/webViewController/interceptors/CardCreationInterceptor.kt
```diff
@@ -0,0 +1,67 @@
...
+    override fun intercept(request: WebResourceRequest): Boolean {
+        val url = request.url.toString()
+
+        if (url.contains("https://api.mercuryo.io/v1.6/cards")) { // Specify your condition here
```
valentunn, line 34: Should we try to be more robust and support increase of the api version? E.g. if mercuryo changes api version to 1.7 this logic wont work, but at least it should try to

feature-assets/src/main/java/io/novafoundation/nova/feature_assets/presentation/novacard/overview/webViewController/interceptors/NovaCardInterceptor.kt
```diff
@@ -0,0 +1,64 @@
...
+
+fun makeRequest(okHttpClient: OkHttpClient, requestBuilder: Request.Builder): Response {
+    val okHttpResponse = okHttpClient.newCall(requestBuilder.build()).execute()
+
+    if (okHttpResponse.isSuccessful) {
+        return okHttpResponse
+    }
+
+    throw RuntimeException("Request failed with ${okHttpResponse.code} code: ${okHttpResponse.networkResponse?.body}")
+}
```
valentunn, line 26: I think it should be an extension for `OkHttpClient`: `OkHttpClient.makeRequestBlocking`

feature-assets/src/main/java/io/novafoundation/nova/feature_assets/presentation/novacard/overview/webViewController/interceptors/TopUpRequestInterceptor.kt
```diff
@@ -0,0 +1,128 @@
...
+
+        fun onTopUpCompleted(orderId: String)
+
+        fun onTopUpStart(orderId: String, amount: BigDecimal, address: String)
+    }
+
+    override fun intercept(request: WebResourceRequest): Boolean {
+        val url = request.url.toString()
+
+        val pattern = Regex("https://api\\.mercuryo\\.io/[a-zA-Z0-9.]+/widget/sell-request/([a-zA-Z0-9]+)/status.*")
```
valentunn, line 33: This is much better approach. I think we should adapt it in `CardCreationInterceptor` as well

feature-assets/src/main/java/io/novafoundation/nova/feature_assets/presentation/novacard/overview/webViewController/interceptors/TopUpRequestInterceptor.kt
```diff
@@ -0,0 +1,128 @@
...
+ *     "data": {
+ *         "status": "completed", // Status may be new, pending, completed
+ *         "is_partially_paid": 0,
+ *         "amounts": {
+ *             "request": {
+ *                 "amount": "3.55",
+ *                 "currency": "DOT",
+ *                 "fiat_amount": "25.00",
+ *                 "fiat_currency": "EUR"
+ *             },
```
valentunn, line 77: Are you sure its needed anymore?

feature-assets/src/main/java/io/novafoundation/nova/feature_assets/presentation/novacard/topup/TopUpCardViewModel.kt
```diff
@@ -0,0 +1,195 @@
...
+
+    private fun setupFees() {
+        feeMixin.connectWith(
+            feeMixin.feeChainAssetFlow,
+            amountChooserMixin.amountState
+        ) { feePaymentCurrency, commissionAsset, amountState ->
+            val assetTransfer = buildTransfer(feePaymentCurrency, amountState.inputKind.isMaxAction())
+
+            val originFee = sendInteractor.getOriginFee(assetTransfer, viewModelScope)
+            originFee.submissionFee
```
valentunn, line 174: Yeah so `OriginFee` is not needed at all. Lets use `SubmissionFee` directy then

feature-assets/src/main/java/io/novafoundation/nova/feature_assets/data/repository/NovaCardStateRepository.kt
antonijzelinskij: Agree. But since retesting cards is too complicated I'd like to don't fix small issues to avoid retest

feature-assets/src/main/java/io/novafoundation/nova/feature_assets/domain/novaCard/NovaCardInteractor.kt
antonijzelinskij: Yes thank you. I forgot to change a variable name after we decide to show top up timer everytime

feature-assets/src/main/java/io/novafoundation/nova/feature_assets/domain/novaCard/NovaCardInteractor.kt
antonijzelinskij: Agree. Fixed

feature-assets/src/main/java/io/novafoundation/nova/feature_assets/presentation/novacard/overview/webViewController/NovaCardWebViewController.kt
antonijzelinskij: I think the both ways is nice, but I still prefer to left it as it is to avoid retest

feature-assets/src/main/java/io/novafoundation/nova/feature_assets/presentation/novacard/overview/webViewController/interceptors/CardCreationInterceptor.kt
antonijzelinskij: Agree. Will do it

feature-assets/src/main/java/io/novafoundation/nova/feature_assets/presentation/novacard/overview/webViewController/interceptors/TopUpRequestInterceptor.kt
antonijzelinskij: Yes, let's have it. It may be useful in future

PR #1869, opened 2025-03-19
Feature/unified address format

antonijzelinskij:


feature-account-api/src/main/java/io/novafoundation/nova/feature_account_api/presenatation/actions/ExternalActions.kt
```diff
@@ -37,20 +41,23 @@ interface ExternalActions : Browserable {
...
+        showAddressActions(chain.accountIdOf(address), chain)
+    }
+}
 
-suspend fun ExternalActions.Presentation.showAddressActions(address: String, chain: Chain) = showExternalActions(
-    type = ExternalActions.Type.Address(address),
+suspend fun ExternalActions.Presentation.showAddressActions(accountId: AccountId, chain: Chain) = showExternalActions(
+    type = ExternalActions.Type.Address(chain.addressOf(accountId), ChainWithAccountId(chain, accountId)),
     chain = chain
 )
```
antonijzelinskij, line 63: Use this approach to avoid huge refactoring

runtime/src/main/java/io/novafoundation/nova/runtime/ext/ChainExt.kt
```diff
@@ -327,7 +339,8 @@ fun Chain.isValidAddress(address: String): Boolean {
         } else {
             address.toAccountId() // verify supplied address can be converted to account id
 
-            address.addressPrefix() == addressPrefix.toShort()
+            addressPrefix.toShort() == address.addressPrefix() ||
+                legacyAddressPrefix?.toShort() == address.addressPrefix()
```
valentunn, line 343: `addressPrefix` can be fetched only once

app/src/main/java/io/novafoundation/nova/app/root/presentation/RootViewModel.kt
```diff
@@ -89,6 +91,8 @@ class RootViewModel(
         syncPushSettingsIfNeeded()
 
         externalServiceInitializer.initialize()
+
+        toastMessageManager.toastMessagesEvents.observeForever { showToast(it.peekContent()) }
```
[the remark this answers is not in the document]
antonijzelinskij: I just used the same way for wallet connect feature that you can find upper.
Okay, I'll send it to RootActivity

app/src/main/res/navigation/split_screen_nav_graph.xml
```diff
@@ -1074,6 +1078,12 @@
         android:id="@+id/dappFavorites"
         android:name="io.novafoundation.nova.feature_dapp_impl.presentation.favorites.DappFavoritesFragment"
         android:label="fragment_favorites"
-        tools:layout="@layout/fragment_favorites_dapp"/>
+        tools:layout="@layout/fragment_favorites_dapp" />
+
+    <dialog
```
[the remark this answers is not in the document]
antonijzelinskij: Yes, you are right.
Moved it to root + added `navigateInRoot()` in router. I think it's easiest way to not duplicate actions

feature-account-api/src/main/java/io/novafoundation/nova/feature_account_api/presenatation/actions/ExternalActionsProvider.kt
```diff
@@ -52,11 +56,13 @@ class ExternalActionsProvider(
 
         // only show icon for address as for now
         val icon = when (type) {
-            is ExternalActions.Type.Address -> type.address?.let { address ->
-                addressIconGenerator.createOptionalAccountAddressIcon(chain, address)
-            } ?: resourceManager.getDrawable(R.drawable.ic_identicon_placeholder)
+            is ExternalActions.Type.Address -> addressIconGenerator.createOptionalAccountAddressIcon(chain, type.address)
```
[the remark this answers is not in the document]
antonijzelinskij: It was called anyway using address?.let {}, so address was not null there.
I found run catching inside so as I see when we use address instead of accountId it may be failed

feature-account-impl/src/main/java/io/novafoundation/nova/feature_account_impl/presentation/common/address/RealCopyAddressMixin.kt
```diff
@@ -0,0 +1,66 @@
...
+    override fun copyLegacyAddress(chainWithAccountId: ChainWithAccountId) {
+        copyAddress(getLegacyAddress(chainWithAccountId))
+    }
+
+    override fun getPrimaryAddress(chainWithAccountId: ChainWithAccountId): String {
+        return chainWithAccountId.chain.addressOf(chainWithAccountId.accountId)
+    }
+
+    override fun getLegacyAddress(chainWithAccountId: ChainWithAccountId): String? {
+        return chainWithAccountId.chain.legacyAddressOfOrNull(chainWithAccountId.accountId)
```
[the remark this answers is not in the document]
antonijzelinskij: I agree that logically this methods shouldn't be public, however, I wanted to accumulate this logic in one place to avoid mistakes when copying the address.
So this Mixin can generate address and can copy that and all in one place.

runtime/src/main/java/io/novafoundation/nova/runtime/ext/ChainExt.kt
antonijzelinskij: Yes but I think it's okay here to keep code shorter: https://refactoring.guru/ru/replace-temp-with-query

valentunn: Main main concern was not readability but performance actually - `addressPrefix` is not super-trivial method under the hood
Anyways this is quite minor

app/src/main/java/io/novafoundation/nova/app/root/navigation/navigators/Navigator.kt
```diff
@@ -482,7 +482,7 @@ class Navigator(
 
         navigationBuilder().action(R.id.action_openUnifiedAddressDialog)
             .setArgs(ChainAddressSelectorFragment.getBundle(payload))
-            .navigateInFirstAttachedContext()
+            .navigateInRoot()
```
valentunn, line 485: Still, we are missing unified "when to use what" guide that will help us to avoid navigation issues entirely. Lets leave a comment for  for `navigationBuilder` and `navigateInFirstAttachedContext` / `navigateInRoot` that will also describe how actions should be added in xcml e.t.c

Like imagine a new person reading this code. They will for sure ask a question "why do we use navigateInRoot here and navigateInFirstAttachedContext in other places?". Given that we didnt manage to create a clear system from the beginning, lets address it with proper docs at the very least

app/src/main/res/navigation/root_nav_graph.xml
```diff
@@ -133,6 +133,10 @@
         android:id="@+id/action_open_select_wallet"
         app:destination="@id/selectWalletFragment" />
 
+    <action
```
valentunn, line 136: If we always use root navigaiton, does it mean that the shown screen will always overlay the minimized browser view?

feature-assets/src/main/java/io/novafoundation/nova/feature_assets/presentation/receive/ReceiveViewModel.kt
```diff
@@ -121,4 +126,12 @@ class ReceiveViewModel(
             tokenType.symbol
         ) + " " + address
     }
+
+    fun openLegacyAddressScreen() {
```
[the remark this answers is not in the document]
antonijzelinskij: Agree, fixed

valentunn (approved):

PR #1887, opened 2025-03-26
Optimize Ledger connection

antonijzelinskij:
This change reduced the query execution time for delegations by 2x.
21s -> 11s


valentunn (approved):

valentunn (commented):
From the conversation with iOS team: we should not hardcode MTU value but rather let device negatiate it with the Ledger

PR #1888, opened 2025-03-28
Fix/dynamic xcm bugs

valentunn:
* Switch to master urls
* Support new version of dry run
* Support system chains teleports
* Detect deposits only to recipient account
* Always rely on forwarded xcms when retrieving xcm
* Construct reserve location separately - do not take it from token location as it may contain extra junctions


feature-wallet-impl/src/main/java/io/novafoundation/nova/feature_wallet_impl/data/repository/RealCrossChainTransfersRepository.kt
```diff
@@ -26,7 +27,8 @@ private const val DYNAMIC_CACHE_NAME = "RealCrossChainTransfersRepository.Dynami
 class RealCrossChainTransfersRepository(
     private val api: CrossChainConfigApi,
     private val fileCache: FileCache,
-    private val gson: Gson
+    private val gson: Gson,
+    private val parachainInfoRepository: ParachainInfoRepository,
```
valentunn, line 31: Not ideal to have repository in repository but I did not quickly came up with a way to improve this without noticable refactoring of upper-layers - either TokenReserveRegistry uses `ParachainInfoRepository` to lazily fetch paraIds to consrtuct `ChainLocation` instances or it is done on layers above (which requires more work)
I will think about it in background. Maybe we can abstract creation of registry with a factory

PR #1891, opened 2025-04-02
Feature/off ramp new UI

antonijzelinskij:


feature-assets/src/main/java/io/novafoundation/nova/feature_assets/presentation/balance/common/buySell/BuySellSelectorMixin.kt
```diff
@@ -0,0 +1,68 @@
...
+import io.novafoundation.nova.feature_assets.presentation.AssetsRouter
+import io.novafoundation.nova.feature_assets.presentation.balance.common.buySell.BuySellSelectorMixin.Selector
+
+interface BuySellSelectorMixin {
+
+    sealed interface Selector {
+
+        object AllAssets : Selector
+
+        class Asset(val chaiId: String, val assetId: Int) : Selector
```
valentunn, line 17: We can use `FullChainAssetId`

feature-assets/src/main/java/io/novafoundation/nova/feature_assets/presentation/balance/detail/BalanceDetailViewModel.kt
```diff
@@ -126,18 +129,20 @@ class BalanceDetailViewModel(
         .inBackground()
         .share()
 
+    val buySellSelectorMixin = buySellSelectorMixinFactory.create()
+
     val chainUI = chainFlow.map { mapChainToUi(it) }
 
-    val buyMixin = buyMixinFactory.create(scope = this)
+    val buyMixin = tradeMixinFactory.create(scope = this)
```
valentunn, line 136: I wonder why we do have multiple mixins for the single feature? At the moment, it is card to understand what each of them does based just on the name, which is not great

feature-buy-api/src/main/java/io/novafoundation/nova/feature_buy_api/domain/TradeTokenRegistry.kt
```diff
@@ -0,0 +1,56 @@
...
+
+        fun createIntegrator(chainAsset: Chain.Asset, address: String, tradeFlow: TradeFlow): I
+    }
+
+    interface Integrator<T> {
+
+        fun openFlow(using: T)
+    }
+
+    enum class TradeFlow {
```
valentunn, line 42: I suggest to rename it to `TradeType` or something, since when I first saw `TradeFlow` name I first though it it a typealias to something like `Flow<Trade>`

feature-buy-api/src/main/java/io/novafoundation/nova/feature_buy_api/domain/TradeTokenRegistry.kt
```diff
@@ -0,0 +1,56 @@
...
+    sealed interface PaymentMethod {
+        object Visa : PaymentMethod
+        object MasterCard : PaymentMethod
+        object ApplePay : PaymentMethod
+        object GooglePay : PaymentMethod
+        object Sepa : PaymentMethod
+        object BankTransfer : PaymentMethod
+
+        class Other(val quantity: Int) : PaymentMethod
+    }
```
valentunn, line 55: I feel like domain-level PaymentMethod should just be the enumeration of all possible types, like Visa, MasterCard, e.t.c.
And `Other` should only appear on presentation level, so in the ViewModel or in mixin

feature-buy-impl/src/main/java/io/novafoundation/nova/feature_buy_impl/domain/providers/BanxaProvider.kt
```diff
@@ -19,9 +20,27 @@ class BanxaProvider(
...
+                TradeTokenRegistry.PaymentMethod.MasterCard,
+                TradeTokenRegistry.PaymentMethod.ApplePay,
+                TradeTokenRegistry.PaymentMethod.GooglePay,
+                TradeTokenRegistry.PaymentMethod.Sepa,
+                TradeTokenRegistry.PaymentMethod.Other(5)
+            )
+
+            TradeTokenRegistry.TradeFlow.SELL -> throw IllegalStateException("Sell is not supported for Banxa provider")
+        }
+    }
```
valentunn, line 41: Minor: its not ideal that we have to handle "sell is not supported" twice. In particular, that we allow to invoke this class with `SELL` even when `SELL` is not supported
This can be solved, for example, by exposing only`getPaymentMethods` and returning empty list for `SELL`

feature-buy-impl/src/main/java/io/novafoundation/nova/feature_buy_impl/domain/providers/MercuryoProvider.kt
```diff
@@ -24,11 +25,34 @@ class MercuryoProvider(
...
+
+    override fun getPaymentMethods(tradeFlow: TradeTokenRegistry.TradeFlow): List<TradeTokenRegistry.PaymentMethod> {
+        return when (tradeFlow) {
+            TradeTokenRegistry.TradeFlow.BUY -> listOf(
+                TradeTokenRegistry.PaymentMethod.Visa,
+                TradeTokenRegistry.PaymentMethod.MasterCard,
+                TradeTokenRegistry.PaymentMethod.ApplePay,
+                TradeTokenRegistry.PaymentMethod.GooglePay,
+                TradeTokenRegistry.PaymentMethod.Sepa,
+                TradeTokenRegistry.PaymentMethod.Other(5)
```
valentunn, line 41: I dont think this should be hard-coded. This was just mockups fake data

feature-assets/src/main/java/io/novafoundation/nova/feature_assets/presentation/balance/detail/BalanceDetailViewModel.kt
antonijzelinskij: I already fixed it in branch with improving buy sell buttons. I'll make pr a bit later

feature-buy-api/src/main/java/io/novafoundation/nova/feature_buy_api/domain/TradeTokenRegistry.kt
antonijzelinskij: Different providers have different "other" payment methods. So I think it's more convenient to keep it here and easier for refactoring

feature-buy-impl/src/main/java/io/novafoundation/nova/feature_buy_impl/domain/providers/MercuryoProvider.kt
antonijzelinskij: Could you please explain what do you mean and what do you suggest?

valentunn: Explicitly discuss the logic here with design team: how many providers should be shown separately e.t.c

feature-buy-api/src/main/java/io/novafoundation/nova/feature_buy_api/domain/TradeTokenRegistry.kt
valentunn: This looks like an assumption incorrectly derived from mockups. As I mentioned below, we should discuss it with Andrey

valentunn (approved):

PR #1894, opened 2025-04-02
Take disabled chains into account in swaps

valentunn:
* SwapService now only invokes single chain factories on enabled chains
* Cross Chain exchange filters out disabled directions


stepanLav:
Task linked: [CU-86975ajt3 Android. Crash on swaps if disable Hydration](https://app.clickup.com/t/86975ajt3)

antonijzelinskij (approved):

PR #1909, opened 2025-04-09
Fix/ledger state machine

valentunn:
* Move work related to permissions to the State machine. For now, just use `permissionsGranted: true/false`
* Ability to initiate discovery for each methods separately, when all relevant requirements has been satisfied
* DiscoveryService is singleton again. This elimintes the issue of multiple discovery services present in the system which increases mental comlexity. It also should fix the crash when multiple transasctions are submitted one by one since logic with subscribptions tracking was not working at all since we were creating a new instance of discovery service each time
* Discovery Service can now start and stop discovery for each method separately
* Construct initial state in the single place, including permissions and requirements checks
* State machine states now support `boostrap` method that will be invoked for the initial state so it can launch some initial side effects or even perform state transiction by itself. So we can make ViewModel as dump as possible and put all logic in state machine
* Human-readable state machine logs
* Explicitly order requirement requests so awaitable (e.g. bluetooth) will be executed before fire-and-forget ones (e.g. Location). This was done implicitly before by just ordering permission in the "right" way


feature-ledger-impl/src/main/java/io/novafoundation/nova/feature_ledger_impl/presentation/account/common/selectLedger/stateMachine/states/SelectLedgerState.kt
```diff
@@ -1,67 +1,170 @@
...
+    protected suspend fun requestPermissions(
+        allDiscoveryMethods: DiscoveryMethods,
+        newRequirementsAvailability: DiscoveryRequirementAvailability,
+        usedAllowedRequirementAvailabilityRequests: Boolean
+    ) {
+        val allDiscoveryRequirements = allDiscoveryMethods.discoveryRequirements()
+
+        // We can only request permissions if there is a single method or a user explicitly pressed a button to request permissions
+        // Otherwise we don't bother them with the automatic requests as there might be method that do not require permissions at all
+        val canRequestPermissions = allDiscoveryMethods.methods.size == 1 || usedAllowedRequirementAvailabilityRequests
```
antonijzelinskij, line 93: As I see we duplicate the method: `canRequestDiscoveryRequirements`. Let's use it here

PR #1913, opened 2025-04-10
Block input view on top up

antonijzelinskij:


common/src/main/java/io/novafoundation/nova/common/view/shape/ShapeProvider.kt
```diff
@@ -69,8 +69,8 @@ fun Context.getInputBackground() = getCornersStateDrawable(
     idleDrawable = getRoundedCornerDrawable(R.color.input_background),
     disabledDrawable = getRoundedCornerDrawableFromColors(
         fillColor = getColor(R.color.input_background),
-        strokeColor = getColor(R.color.container_border),
-        strokeSizeInDp = 1f
+        strokeColor = null,
+        strokeSizeInDp = 0f
     )
```
antonijzelinskij, line 74: Removed a border for disabled input state to consider design

valentunn (approved):

PR #1914, opened 2025-04-10
Chore/refactor signers

valentunn:
Requires https://github.com/novasamatech/substrate-sdk-android/pull/104


runtime/src/main/java/io/novafoundation/nova/runtime/extrinsic/CustomTransactionExtensions.kt
```diff
@@ -0,0 +1,21 @@
...
+        defaultValues().forEach(::setTransactionExtension)
+    }
+
+    fun defaultValues(): List<TransactionExtension> {
+        return listOf(
+            ChargeAssetTxPayment(),
+            CheckAppId()
+        )
+    }
+}
```
antonijzelinskij, line 21: Maybe it's better to use as extension methods, because calling CustomTransactionExtensions.applyDefaultValues() in apply may confuse

antonijzelinskij (approved):

PR #1915, opened 2025-04-11
Show full amount on asset details

antonijzelinskij:

### DIFF
--- a/common/src/main/java/io/novafoundation/nova/common/utils/TokenSymbol.kt
+++ b/common/src/main/java/io/novafoundation/nova/common/utils/TokenSymbol.kt
@@ -21,18 +21,8 @@ fun BigDecimal.formatTokenAmount(roundingMode: RoundingMode = RoundingMode.FLOOR
     return format(roundingMode)
 }
 
-fun BigDecimal.formatTokenAmount(
-    tokenSymbol: TokenSymbol,
-    roundingMode: RoundingMode = RoundingMode.FLOOR,
-    includeAssetTicker: Boolean = true
-): String {
-    val formatted = format(roundingMode)
-
-    return if (includeAssetTicker) {
-        formatted.withTokenSymbol(tokenSymbol)
-    } else {
-        formatted
-    }
+fun BigDecimal.formatTokenAmount(tokenSymbol: TokenSymbol, roundingMode: RoundingMode = RoundingMode.FLOOR): String {
+    return format(roundingMode).withTokenSymbol(tokenSymbol)
 }
 
 fun String.withTokenSymbol(tokenSymbol: TokenSymbol): String {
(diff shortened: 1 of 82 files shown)


feature-wallet-api/src/main/res/layout/view_balances.xml
```diff
@@ -26,17 +26,24 @@
...
             android:layout_height="wrap_content"
             android:layout_marginStart="16dp"
             android:layout_marginTop="8dp"
+            android:layout_marginEnd="16dp"
+            android:ellipsize="end"
+            android:lines="1"
             android:textColor="@color/text_primary"
+            app:autoSizeMaxTextSize="22sp"
+            app:autoSizeMinTextSize="14sp"
+            app:autoSizeTextType="uniform"
```
antonijzelinskij, line 42: Made adaptive text size for amount view

common/src/main/java/io/novafoundation/nova/common/utils/TokenSymbol.kt
```diff
@@ -21,8 +21,18 @@ fun BigDecimal.formatTokenAmount(roundingMode: RoundingMode = RoundingMode.FLOOR
     return format(roundingMode)
 }
 
-fun BigDecimal.formatTokenAmount(tokenSymbol: TokenSymbol, roundingMode: RoundingMode = RoundingMode.FLOOR): String {
-    return format(roundingMode).withTokenSymbol(tokenSymbol)
+fun BigDecimal.formatTokenAmount(
+    tokenSymbol: TokenSymbol,
+    roundingMode: RoundingMode = RoundingMode.FLOOR,
+    includeAssetTicker: Boolean = true
```
valentunn, line 27: This seems redundant as we can just use `format(roundingMode)` directly

feature-wallet-api/src/main/java/io/novafoundation/nova/feature_wallet_api/presentation/model/AmountModel.kt
```diff
@@ -59,16 +62,23 @@ fun mapAmountToAmountModel(
...
+        amount.formatTokenAmount(token.configuration, roundingMode, includeAssetTicker)
     } else {
-        amount.format(roundingMode)
+        val unformattedAmount = amount.format(roundingMode)
+
+        if (includeAssetTicker) {
+            unformattedAmount.withTokenSymbol(token.configuration.symbol)
+        } else {
+            unformattedAmount
+        }
```
valentunn, line 81: I do not quite get the logic here: `formatTokenAmount` calls `amount.format(roundingMode)` anyway so there is no effective difference between these if branches?

Also the task was to show **full** amount, meaning that we should format with the formatter that doesnt have abbreviations enabled

valentunn:
Logic in screenshot is incorrect. The only thing that we should disable are abbreviations. We should not show 10+ decimal places still

PR #1916, opened 2025-04-11
Support sell providers from nova utils

antonijzelinskij:


common/src/main/java/io/novafoundation/nova/common/utils/KotlinExt.kt
```diff
@@ -690,3 +690,7 @@ suspend fun <T> scopeAsync(
 fun Int.collectionIndexOrNull(): Int? {
     return takeIf { it >= 0 }
 }
+
+fun <T> Set<T>.intersectsWith(other: Set<T>): Boolean {
+    return this.any { it in other }
+}
```
valentunn, line 696: I'd name it `hasIntersectionWith` as `intersectsWith` is a little bit vague to me

valentunn (approved):

PR #1927, opened 2025-04-17
Use dapp known icons as default icon in tabs fragment

antonijzelinskij:


feature-dapp-impl/src/main/java/io/novafoundation/nova/feature_dapp_impl/presentation/tab/BrowserTabRvItem.kt
```diff
@@ -3,6 +3,7 @@ package io.novafoundation.nova.feature_dapp_impl.presentation.tab
 data class BrowserTabRvItem(
     val tabId: String,
     val tabName: String?,
+    val knownDappIconUrl: String?,
     val tabFaviconPath: String?,
```
valentunn, line 7: I think its better to use `Icon` here and add `FromFile` subclass to `Icon`. This way we wont need two separate fields and extra logic on ui level

valentunn (approved):

feature-dapp-impl/src/main/java/io/novafoundation/nova/feature_dapp_impl/presentation/tab/BrowserTabRvItem.kt
antonijzelinskij: Yes it's more clear, thank you

PR #1929, opened 2025-04-22
Wallet connect pairing dapp handling

antonijzelinskij:
#8698hzzv8

### DIFF
--- a/feature-dapp-impl/src/main/java/io/novafoundation/nova/feature_dapp_impl/web3/webview/interceptors/WalletConnectPairingInterceptor.kt
+++ b/feature-dapp-impl/src/main/java/io/novafoundation/nova/feature_dapp_impl/web3/webview/interceptors/WalletConnectPairingInterceptor.kt
@@ -1,0 +1,28 @@
+package io.novafoundation.nova.feature_dapp_impl.web3.webview.interceptors
+
+import android.net.Uri
+import android.webkit.WebResourceRequest
+import io.novafoundation.nova.common.interfaces.ActivityIntentProvider
+import io.novafoundation.nova.common.resources.ContextManager
+import io.novafoundation.nova.common.utils.webView.WebViewRequestInterceptor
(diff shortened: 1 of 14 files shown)


stepanLav:
Task linked: [CU-8698hzzv8 Android. DApps. Improve Deeplinks handling](https://app.clickup.com/t/8698hzzv8)

feature-dapp-impl/src/main/java/io/novafoundation/nova/feature_dapp_impl/web3/webview/interceptors/WalletConnectPairingInterceptor.kt
```diff
@@ -0,0 +1,28 @@
...
+
+class WalletConnectPairingInterceptor(
+    private val contextManager: ContextManager,
+    private val activityIntentProvider: ActivityIntentProvider
+) : WebViewRequestInterceptor {
+
+    override fun intercept(request: WebResourceRequest): Boolean {
+        if (WalletConnectUtils.isWalletConnectPairingLink(request.url)) {
+            runIntent(request.url)
+            return true
```
valentunn, line 18: So this is no different from the previous logic - it will still show the system chooser in case there are multipole activities matching given one. The idea was to invoke logic of `WalletConnectPairDeeplinkHandler` directly when we detect the relevant deeplink

antonijzelinskij: But maybe we need to run this directly and it will be more clear:
`walletConnectService.pair(data.toString())`

valentunn: > But maybe we need to run this directly and it will be more clear: `walletConnectService.pair(data.toString())`

Yep I think this is better

PR #1931, opened 2025-04-25
Updated App bar UI

antonijzelinskij:


feature-assets/src/main/java/io/novafoundation/nova/feature_assets/presentation/balance/list/BalanceListViewModel.kt
```diff
@@ -325,4 +312,7 @@ class BalanceListViewModel(
     fun switchViewMode() {
         launch { assetListMixin.switchViewMode() }
     }
+
+    fun settingsClicked() {
```
antonijzelinskij, line 316: Going to handle it in pr with new tab

valentunn (approved):

PR #1940, opened 2025-05-08
Feature/shop auth

antonijzelinskij:
#8698v04py


bindings/merlin/rust/src/lib.rs
```diff
@@ -0,0 +1,111 @@
...
+//     let public = private.to_public();
+
+//     let nonce = decode_hex("2e7857995eedb9ffc3f6930cfea9765dbe712a667e4158b17f7a11ae4a39351b").expect("Cant fail");
+
+//     let mut transcript = Transcript::new(b"raise-auth");
+//     transcript.append_message(b"challenge-nonce", &nonce);
+
+//     let signature = private.sign(transcript, &public);
+//     println!("{}", encode_hex(signature.to_bytes().as_slice()));
+// }
```
antonijzelinskij, line 111: I completely copied it from polkadot app but I guess we can remove it?

feature-pay-impl/src/main/java/io/novafoundation/nova/feature_pay_impl/data/raise/auth/network/RaiseAuthInterceptor.kt
```diff
@@ -0,0 +1,247 @@
...
+    @Throws(IOException::class)
+    override fun intercept(chain: Interceptor.Chain): Response {
+        var request: Request = chain.request()
+
+        if (RaiseEndpoints.Auth.isAuthEndpoint(request.url.toString())) return chain.proceed(request)
+
+        val builder: Request.Builder = request.newBuilder()
+        builder.header("Accept", "application/json")
+
+        val selectedMetaAccount = runBlocking { accountRepository.getSelectedMetaAccount() }
```
antonijzelinskij, line 48: In Nova case we should keep meta account in interceptor to avoid problems with changing account during refresh token

stepanLav:
Task linked: [CU-8698v04py Raise API Integration - Auth](https://app.clickup.com/t/8698v04py)

bindings/merlin/rust/src/lib.rs
valentunn: Yep this can be removed entirely. I used it to test rust code separately

feature-account-api/src/main/java/io/novafoundation/nova/feature_account_api/data/secrets/AccountSecretsFactory.kt
```diff
@@ -0,0 +1,39 @@
+package io.novafoundation.nova.feature_account_api.data.secrets
+
+import io.novafoundation.nova.common.data.secrets.v2.ChainAccountSecrets
+import io.novafoundation.nova.common.data.secrets.v2.MetaAccountSecrets
+import io.novafoundation.nova.core.model.CryptoType
+import io.novasama.substrate_sdk_android.scale.EncodableStruct
+import io.novasama.substrate_sdk_android.scale.Schema
+
+interface AccountSecretsFactory {
```
valentunn, line 9: As we discussed on the planning, we should sign the challenge with actual user keypair as it is safe to do because we use different sr25519 context (under the hood in merlin rust module)
So nothing additional should be exposed from the account module, but instead we should do the following:

1. Fetch account keypair via `SecretStoreV2.getAccountSecrets().keypair(polkadot)` (also sync on this with Ruslan)
2. Check that the returned keypair is Sr25519 keypair
3. Sign with the found keypair

feature-pay-impl/src/main/java/io/novafoundation/nova/feature_pay_impl/data/raise/brands/ShopBrandsRepository.kt
```diff
@@ -0,0 +1,66 @@
...
+private const val BRANDS_PAGE_SIZE = 50
+private const val EMPTY_QUERY = ""
+
+class RealShopBrandsRepository(
+    private val raiseBrandsApi: RaiseBrandsApi,
+    private val networkStateService: NetworkStateService,
+    private val raiseBrandsConverter: RaiseBrandsConverter
+) : ShopBrandsRepository {
+
+    private val firstPageBrands: MutableSharedFlow<DataPage<RaiseBrand>> = singleReplaySharedFlow()
```
valentunn, line 36: If we will remove prefetch, do not forget to fill this flow in `getBrands`, otherwise first page would never load

valentunn (changes requested):

feature-pay-impl/src/main/java/io/novafoundation/nova/feature_pay_impl/data/raise/brands/ShopBrandsRepository.kt
antonijzelinskij: Okay, already did it in next pr

feature-account-api/src/main/java/io/novafoundation/nova/feature_account_api/data/secrets/AccountSecretsFactory.kt
antonijzelinskij: As we discussed with Ruslan:
- We use seed from secrets
- Then we create a secret using this seed and a derivation path
- Remove SR25119 requirement, since we always create a keypair using seed

feature-pay-impl/src/main/java/io/novafoundation/nova/feature_pay_impl/domain/Utils.kt
```diff
@@ -1,9 +1,8 @@
 package io.novafoundation.nova.feature_pay_impl.domain
 
-import io.novafoundation.nova.core.model.CryptoType
 import io.novafoundation.nova.feature_account_api.domain.model.LightMetaAccount
 import io.novafoundation.nova.feature_account_api.domain.model.MetaAccount
 
 fun MetaAccount.isValidForShopApi(): Boolean {
-    return type == LightMetaAccount.Type.SECRETS && substrateCryptoType == CryptoType.SR25519
+    return type == LightMetaAccount.Type.SECRETS
```
antonijzelinskij, line 7: Since we create a keypair from seed I deleted requirement for SR25519 crypto type

valentunn: Looks like this is still not sufficient as we dont always have a passphrase for SECRETS account. For example, account imported via JSON file wont have it

valentunn (approved):

PR #1965, opened 2025-06-01
Fixed bug with waiting dialog open for nova card

antonijzelinskij:
#86998e8df

### DIFF
--- a/feature-assets/src/main/java/io/novafoundation/nova/feature_assets/presentation/novacard/overview/NovaCardViewModel.kt
+++ b/feature-assets/src/main/java/io/novafoundation/nova/feature_assets/presentation/novacard/overview/NovaCardViewModel.kt
@@ -28,9 +28,11 @@ import io.novafoundation.nova.runtime.multiNetwork.chain.model.Chain
 import kotlinx.coroutines.flow.first
 import kotlinx.coroutines.launch
 import java.math.BigDecimal
+import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.flow.launchIn
 import kotlinx.coroutines.flow.map
 import kotlinx.coroutines.flow.onEach
+import kotlinx.coroutines.withContext
 
 class NovaCardViewModel(
     private val chainRegistry: ChainRegistry,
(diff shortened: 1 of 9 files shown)


stepanLav:
Task linked: [CU-86998e8df Android. Mercuryo - Top-up BS doesn't opened](https://app.clickup.com/t/86998e8df)

feature-assets/src/main/java/io/novafoundation/nova/feature_assets/presentation/novacard/overview/NovaCardViewModel.kt
```diff
@@ -121,11 +127,11 @@ class NovaCardViewModel(
...
             .onEach {
                 when (it) {
-                    TopUpAddressResponder.Response.Cancel -> {
+                    TopUpAddressResponder.Response.Cancel -> withContext(Dispatchers.Main) {
                         assetsRouter.returnToMainScreen()
                     }
 
-                    TopUpAddressResponder.Response.Success -> {
+                    TopUpAddressResponder.Response.Success -> withContext(Dispatchers.Main) {
                         updateCardState()
```
antonijzelinskij, line 135: That's lines fixed an issue with not opening dialog. I don't know why because this code must be called in main thread and as I checked it was in the main thread. So if you have any idea why this lines works please let me know

valentunn: Yeah this looks very strange. I dont have an idea what might be wrong here. Let's leave a comment that explains why we are using  withContext(Dispatchers.Main)  with a TODO mark to investigate that

feature-assets/src/main/java/io/novafoundation/nova/feature_assets/presentation/novacard/waiting/WaitingNovaCardTopUpFragment.kt
```diff
@@ -14,10 +15,15 @@ class WaitingNovaCardTopUpFragment : BaseBottomSheetFragment<WaitingNovaCardTopU
 
     override fun initViews() {
         dialog?.setCanceledOnTouchOutside(false)
-        isCancelable = false
         getBehaviour().isHideable = false
     }
 
+    override fun onCancel(dialog: DialogInterface) {
+        viewModel.closeNovaCard()
```
valentunn, line 22: Fragment should not command ViewModel but rather notify it about events. I think we should rename viewmodel's method to smth like "onScreenCancelled`

valentunn (approved):

PR #1966, opened 2025-06-01
Added nova card deep link support

antonijzelinskij:
#86998a0zz


valentunn (approved):

feature-assets/src/main/java/io/novafoundation/nova/feature_assets/presentation/novacard/overview/deeplink/NovaCardDeepLinkHandler.kt
```diff
@@ -0,0 +1,38 @@
...
+    override val callbackFlow = MutableSharedFlow<CallbackEvent>()
+
+    override suspend fun matches(data: Uri): Boolean {
+        val path = data.path ?: return false
+
+        return path.startsWith(PATH)
+    }
+
+    override suspend fun handleDeepLink(data: Uri): Result<Unit> = runCatching {
+        automaticInteractionGate.awaitInteractionAllowed()
```
valentunn, line 30: We should probably also await untill account is created?

PR #1967, opened 2025-06-01
Fixed migarion link

antonijzelinskij:
Fixed migration link to iOS supported format

### DIFF
--- a/feature-deep-linking/src/main/java/io/novafoundation/nova/feature_deep_linking/presentation/handling/branchIo/BranchIOConstants.kt
+++ b/feature-deep-linking/src/main/java/io/novafoundation/nova/feature_deep_linking/presentation/handling/branchIo/BranchIOConstants.kt
@@ -4,5 +4,4 @@ object BranchIOConstants {
     const val ACTION_QUERY = "action"
     const val SCREEN_QUERY = "screen"
     const val ENTITY_QUERY = "entity"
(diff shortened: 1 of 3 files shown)


feature-deep-linking/src/main/java/io/novafoundation/nova/feature_deep_linking/presentation/handling/branchIo/BranchIOConstants.kt
```diff
@@ -4,5 +4,4 @@ object BranchIOConstants {
     const val ACTION_QUERY = "action"
     const val SCREEN_QUERY = "screen"
     const val ENTITY_QUERY = "entity"
-    const val SCHEME_QUERY = "scheme"
```
valentunn, line 7: Why did we remove this though?

valentunn (approved):

feature-deep-linking/src/main/java/io/novafoundation/nova/feature_deep_linking/presentation/handling/branchIo/BranchIOConstants.kt
antonijzelinskij: Because I misunderstood Ruslan and I thought that we put scheme in deep link path, but it was wrong since we put it to the query. So we don't need handle scheme in branch io links converter

PR #1969, opened 2025-06-02
Add Unique Network NFT provider

ashkuc:


feature-nft-impl/src/main/java/io/novafoundation/nova/feature_nft_impl/data/source/providers/unique_network/network/response/UniqueNetworkCollection.kt
```diff
@@ -0,0 +1,17 @@
...
+    val coverImage: CoverImage?,
+    val lastTokenId: Int?,
+    val limits: Limits?
+) {
+    data class CoverImage(
+        val url: String?
+    )
+
+    data class Limits(
+        val token_limit: Int?,
```
valentunn, line 15: We can use `SerializedName` annotation to specify snake case names, so the property name can remain in snake case

feature-nft-impl/src/main/java/io/novafoundation/nova/feature_nft_impl/data/source/providers/unique_network/UniqueNetworkProvider.kt
```diff
@@ -0,0 +1,124 @@
...
+                    issuanceMyAmount = null,
+                    price = null,
+                    pricedUnits = null
+                )
+            }
+
+            allNfts += pageNfts
+            offset += pageSize
+
+            if (allNfts.size >= page.count) break
```
valentunn, line 69: Does `page.count` here refer to a total number of NFTs of the user?

valentunn (commented):
Thank you for the contribution! The PR looks good overall. Added a few small suggestions.
Before we can proceed with merging and testing it, could you please attach a few screenshots to showcase the feature?
Also, please provide 1-2 account addresses, where we can test this feature during QA phase

feature-nft-impl/src/main/java/io/novafoundation/nova/feature_nft_impl/data/source/providers/unique_network/network/response/UniqueNetworkCollection.kt
ashkuc: done

feature-nft-impl/src/main/java/io/novafoundation/nova/feature_nft_impl/data/source/providers/unique_network/UniqueNetworkProvider.kt
ashkuc: Yes, page.count refers to the total number of entities matching the query.
In this case, it indicates the total number of NFTs owned by the address on the network.

valentunn (approved):
Thanks you! The PR looks good now. We have a release scheduled for the next week that we want to limit scope of, so we will include this PR into the next one

PR #1970, opened 2025-06-02
fix: correct issuance condition — treat collections with max > 0 as Limited

ashkuc:
Currently, **only** NFT collections with **issuanceTotal = 0** are treated as **Limited**, which seems incorrect.

Please let me know if I'm wrong and there's a reason for this logic.


valentunn (approved):

valentunn:
Thanks you!

PR #1973, opened 2025-06-04
Fixed crash on open new tab in browser

antonijzelinskij:
#86997tv0a


stepanLav:
Task linked: [CU-86997tv0a Android. Crash if open several times one applink with dapp open url](https://app.clickup.com/t/86997tv0a)

feature-dapp-impl/src/main/java/io/novafoundation/nova/feature_dapp_impl/presentation/browser/main/deeplink/DAppDeepLinkHandler.kt
```diff
@@ -36,7 +40,11 @@ class DAppDeepLinkHandler(
 
         ensureDAppInCatalog(normalizedUrl)
 
-        router.openDAppBrowser(DAppBrowserPayload.Address(url))
+        if (browserTabService.hasSelectedTab()) {
+            browserTabService.createAndSelectTab(normalizedUrl)
```
valentunn, line 44: Does this perform navigation to the browser? If not, seems like `router.openDAppBrowser(DAppBrowserPayload.Address(url))` should be moved out of `if` statement

PR #1974, opened 2025-06-05
Fix token assets group ui update

antonijzelinskij:
#869958hkx


stepanLav:
Task linked: [CU-869958hkx Android. Underlay is not displayed after account creation/import](https://app.clickup.com/t/869958hkx)

valentunn (approved):

PR #2006, opened 2025-06-24
Update MS entry point

antonijzelinskij:
#8699b33hx


common/src/main/java/io/novafoundation/nova/common/view/TableView.kt
```diff
@@ -90,6 +101,16 @@ open class TableView @JvmOverloads constructor(
         }
     }
 
+    private fun shouldDrawDivider(child: View) = when {
+        child is TableItem && !child.shouldDrawDivider() -> false
+        child is TableItem && child.shouldDrawDivider() -> true
+        else -> drawDividers
+    }
+
+    fun invalidateChildrenVisibility() {
```
valentunn, line 110: Is it actually needed? We alerady react to visibility changes in `onLayout`

valentunn (approved):

common/src/main/java/io/novafoundation/nova/common/view/TableView.kt
antonijzelinskij: Yes, it's needed in this case
The problem I solved - `onLayout` doesn't called when view is gone

PR #2007, opened 2025-06-24
Multisig/threshold 1

valentunn:
* Support threshold 1 multisig
* New balances validation approach (+ applied to multisig fee & deposit validations)
* Do not sync multlisigs with invalid threshold

### DIFF
--- a/common/src/main/java/io/novafoundation/nova/common/domain/balance/EDCountingMode.kt
+++ b/common/src/main/java/io/novafoundation/nova/common/domain/balance/EDCountingMode.kt
@@ -15,7 +15,7 @@ fun EDCountingMode.calculateBalanceCountedTowardsEd(free: BigInteger, reserved: 
 }
 
 fun EDCountingMode.reservedPreventsDusting(reserved: BigInteger): Boolean {
-    return when(this) {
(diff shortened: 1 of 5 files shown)


feature-account-api/src/main/java/io/novafoundation/nova/feature_account_api/domain/model/MetaAccount.kt
```diff
@@ -157,6 +157,10 @@ fun MultisigAvailability.singleChainId(): ChainId? {
     }
 }
 
+fun MultisigMetaAccount.isThreshold1(): Boolean {
```
antonijzelinskij, line 160: Can we rename it to `isSingleThreshold` to make it more readable?

antonijzelinskij (approved):

feature-account-api/src/main/java/io/novafoundation/nova/feature_account_api/domain/model/MetaAccount.kt
valentunn: I dont think `single` is the right word here and that it would actually make it more understandable

PR #2016, opened 2025-06-30
Poc of signing via reject in dapp browser

valentunn:


valentunn:
Closed as the blocking issue is fixed in polkadot-js/api

PR #2018, opened 2025-07-01
Fixed crash on ms operation for eth-based chain

antonijzelinskij:


feature-account-impl/src/main/java/io/novafoundation/nova/feature_account_impl/presentation/account/wallet/WalletUiUseCaseImpl.kt
```diff
@@ -137,4 +140,12 @@ class WalletUiUseCaseImpl(
...
         return walletSeed(substrateAccountId, ethereumAddress, chainAccounts.values.map(ChainAccount::accountId))
     }
+
+    override suspend fun walletAddressModel(metaAccount: MetaAccount, chain: Chain, iconSize: Int): AddressModel {
+        return AddressModel(
+            address = metaAccount.requireAddressIn(chain),
+            image = walletIcon(metaAccount, iconSize, transparentBackground = true),
+            name = metaAccount.name
+        )
+    }
```
antonijzelinskij, line 150: Used this approach to have Meta-account icon but chain-address

PR #2025, opened 2025-07-03
Add still pending call validation

valentunn:
image


feature-multisig/operations/src/main/java/io/novafoundation/nova/feature_multisig_operations/presentation/details/MultisigOperationDetailsViewModel.kt
```diff
@@ -283,7 +283,7 @@ class MultisigOperationDetailsViewModel(
             fee = feeLoaderMixin.awaitFee(),
             signatoryBalance = signatoryBalance,
             signatory = signatory,
-            chain = chain
+            operation = operationFlow.first()
```
antonijzelinskij, line 286: We can reuse `operation` value that you received before

antonijzelinskij (approved):

PR #2038, opened 2025-07-08
Fixed sell tokens and nova card restriction issues

antonijzelinskij:
- Fixed nova card availability for proxied without polkadot account
- Fixed availability for multisig wallet with threshold1 in nova card and sell. chain account also considered in this case


feature-assets/src/main/java/io/novafoundation/nova/feature_assets/presentation/novacard/common/NovaCardRestrictionCheckMixin.kt
```diff
@@ -0,0 +1,73 @@
...
+
+class NovaCardRestrictionCheckMixin(
+    private val accountUseCase: SelectedAccountUseCase,
+    private val resourceManager: ResourceManager,
+    private val actionLauncher: ActionBottomSheetLauncher,
+) : RestrictionCheckMixin {
+
+    override suspend fun isRestricted(): Boolean {
+        val selectedAccount = accountUseCase.getSelectedMetaAccount()
+        return isRestrictedByMultisig(selectedAccount) || isRestrictedByProxied(selectedAccount)
```
valentunn, line 27: I this the more structued approach would be to do exchaustive when:

``
fun isRestricted(metaAccount) = when(metaAccount) {
 is Multisig -> {..}
 is Proxied -> {..}
is ... -> false
}

This way we wont forget about this mixin when adding a new wallet type
And also we wont need to do additional `        if (metaAccount !is MultisigMetaAccount) return false
` checks inside each function

feature-assets/src/main/java/io/novafoundation/nova/feature_assets/presentation/novacard/common/NovaCardRestrictionCheckMixin.kt
```diff
@@ -0,0 +1,73 @@
...
+            else -> action()
+        }
+    }
+
+    private fun isRestrictedByMultisig(metaAccount: MetaAccount): Boolean {
+        if (metaAccount !is MultisigMetaAccount) return false
+
+        val isThresholdNot1 = !metaAccount.isThreshold1()
+        val chainNotAvailable = !metaAccount.hasChainAccountIn(NOVA_CARD_AVAILABLE_CHAIN_ID)
+        return isThresholdNot1 || chainNotAvailable
```
valentunn, line 45: I think it would be easier to unsderstand if we do]


```
isAllowed = isThreshold1 && hasAccount
return !isAllowed
```

feature-assets/src/main/java/io/novafoundation/nova/feature_assets/presentation/novacard/common/NovaCardRestrictionCheckMixin.kt
```diff
@@ -0,0 +1,73 @@
...
+            isRestrictedByProxied(selectedAccount) -> showProxiedWarning()
+            else -> action()
+        }
+    }
+
+    private fun isRestrictedByMultisig(metaAccount: MetaAccount): Boolean {
+        if (metaAccount !is MultisigMetaAccount) return false
+
+        val isThresholdNot1 = !metaAccount.isThreshold1()
+        val chainNotAvailable = !metaAccount.hasChainAccountIn(NOVA_CARD_AVAILABLE_CHAIN_ID)
```
valentunn, line 44: We do not want to check whether meta account has chain accout, but want to check whether there is an account in such chain in general, so we should use `hasAccountIn`
Using `hasChainAccount` implies the knowledge of internal logic behind `MultisigMetaAccount`, in particular that threhsold 1 multisig accounts have one **chain** account. This knowledge should not be exposed

feature-assets/src/main/java/io/novafoundation/nova/feature_assets/presentation/novacard/common/NovaCardRestrictionCheckMixin.kt
```diff
@@ -0,0 +1,73 @@
...
+        return isThresholdNot1 || chainNotAvailable
+    }
+
+    private fun isRestrictedByProxied(metaAccount: MetaAccount): Boolean {
+        if (metaAccount !is ProxiedMetaAccount) return false
+
+        return !metaAccount.hasChainAccountIn(NOVA_CARD_AVAILABLE_CHAIN_ID)
```
valentunn, line 51: Same here

feature-account-api/src/main/java/io/novafoundation/nova/feature_account_api/domain/model/MetaAccount.kt
```diff
@@ -252,11 +251,27 @@ fun LightMetaAccount.Type.requestedAccountPaysFees(): Boolean {
...
 
-val LightMetaAccount.Type.isProxied: Boolean
-    get() = this == LightMetaAccount.Type.PROXIED
+@OptIn(ExperimentalContracts::class)
+fun LightMetaAccount.isProxied(): Boolean {
+    contract {
+        returns(true) implies (this@isProxied is ProxiedMetaAccount)
+    }
+
+    return type == LightMetaAccount.Type.PROXIED
```
valentunn, line 260: It is better to ensure condition in contract matches the condition in the function itself. Otherwise we might face a crash if type is Proxied but account is not ProxiedMetaAccount (this could be due to some error in the account mapping logic for example)

valentunn (approved):

PR #2039, opened 2025-07-08
Fix and cleanup different nested signer cases

valentunn:
* Fix wrapping order for ProxiedSigner
* Always use proxy/proxied explicit namings
* Unify namings in multisig validations
* Remove unused field in Proxy class
* Use stricter types when it comes to Proxied and Multisig meta accounts
* Fix condition for performing proxy validation - it should be done not when signer is root, but when the delegate signer is the leaf


valentunn:
Superseded by #2040

PR #2041, opened 2025-07-09
Add identity to signatories

antonijzelinskij:
#8699kuz70

### DIFF
--- a/feature-multisig/operations/src/main/java/io/novafoundation/nova/feature_multisig_operations/presentation/details/general/SignatoryListFormatter.kt
+++ b/feature-multisig/operations/src/main/java/io/novafoundation/nova/feature_multisig_operations/presentation/details/general/SignatoryListFormatter.kt
@@ -1,6 +1,7 @@
 package io.novafoundation.nova.feature_multisig_operations.presentation.details.general
 
 import io.novafoundation.nova.common.address.AccountIdKey
+import io.novafoundation.nova.common.list.GroupedList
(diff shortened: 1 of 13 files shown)


feature-multisig/operations/src/main/java/io/novafoundation/nova/feature_multisig_operations/presentation/details/general/SignatoryListFormatter.kt
```diff
@@ -102,9 +79,30 @@ class SignatoryListFormatter(
...
     }
 }
+
+/**
+ * Since we may have multiple accounts by one account id it would be better to show the most relevant to user.
+ * For example we show secrets-account instead of the proxied-account
+ */
+private fun Collection<MetaAccount>.findRelevantAccountToShow(): MetaAccount? {
+    return minByOrNull { it.priorityToShowAsSignatory() }
+}
```
valentunn, line 95: This seems reasonable. I think we should highlight it to iOS
Also, for the signatory that is "our"  signatory I think we should always use it regardless of its wallet type. We can easily do that by integrating logic "if meta id is equal to parentId of current multisig account - use the highest priority"

antonijzelinskij: Agree. Fixed

valentunn (approved):

PR #2045, opened 2025-07-11
Fixed crash on ms transaction create

antonijzelinskij:
#8699m9zwz
#8699r7cd3

### DIFF
--- a/common/src/main/java/io/novafoundation/nova/common/utils/coroutines/RootScope.kt
+++ b/common/src/main/java/io/novafoundation/nova/common/utils/coroutines/RootScope.kt
@@ -3,5 +3,5 @@ package io.novafoundation.nova.common.utils.coroutines
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.MainScope
 
-@Deprecated("Use scope in RootViewModel instead")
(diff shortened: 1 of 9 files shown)


stepanLav:
Task linked: [CU-8699m9zwz Android. Crash on Send ms transaction](https://app.clickup.com/t/8699m9zwz)
Task linked: [CU-8699r7cd3 Android. Staking. Crash on return from background](https://app.clickup.com/t/8699r7cd3)

valentunn (approved):

common/src/main/java/io/novafoundation/nova/common/utils/coroutines/RootScope.kt
```diff
@@ -3,4 +3,5 @@ package io.novafoundation.nova.common.utils.coroutines
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.MainScope
 
+@Deprecated("Use scope in RootViewModel instead")
```
valentunn, line 6: I think it would also be great to explain deprecation so we wont forget about reasons. Something like 
"Using RootScope might have unintended side effects in case RootViewModel, where most of the clients of RootScope are used, is re-created - RootScope jobs wont be cancelled leading to potentially duplicated jobs"

valentunn (approved):

PR #2051, opened 2025-07-15
Android api 35 migration

antonijzelinskij:
#8699tbpx5


stepanLav:
Task linked: [CU-8699tbpx5 Android. Android API 36 Support](https://app.clickup.com/t/8699tbpx5)

feature-versions-impl/src/main/java/io/novafoundation/nova/feature_versions_impl/data/LocalAppVersionProvider.kt
```diff
@@ -19,6 +19,6 @@ class PackageManagerAppVersionProvider(
         } else {
             context.packageManager.getPackageInfo(context.packageName, 0)
         }
-        return packageInfo.versionName
+        return packageInfo.versionName!!
```
valentunn, line 22: It seems like we have two completely dublicate `AppVersionProvider`. Maybe we can remove one of them?

feature-versions-impl/src/main/java/io/novafoundation/nova/feature_versions_impl/data/VersionService.kt
```diff
@@ -96,7 +96,7 @@ class VersionService(
         } else {
             context.packageManager.getPackageInfo(context.packageName, 0)
         }
-        return packageInfo.versionName.toVersion()
+        return packageInfo.versionName!!.toVersion()
```
valentunn, line 99: I suggest to use AppVersionProvider here

PR #2054, opened 2025-07-21
Switching wallet notification

antonijzelinskij:
When you sign transaction using proxied wallet with multisig proxy we switch wallet to multisig.
To improve this UX we notify user about this switching
#8699hfgv2
image


common/src/main/java/io/novafoundation/nova/common/view/bottomSheet/action/UtilsExt.kt
```diff
@@ -43,6 +46,14 @@ fun BottomSheetActionBinding.setupView(
         }
     } ?: neutralButton.makeGone()
 
+    payload.alertModel?.let {
+        alert.makeVisible()
+        alert.apply {
```
mikhail98, line 51: makeVisible() could also be moved inside apply {} block

mikhail98 (approved):
Great!

PR #2061, opened 2025-08-01
Rc/10.1.5

antonijzelinskij:
#8699zkbej


stepanLav:
Task linked: [CU-8699zkbej Android. Nova Shots. Integrity Check Integration](https://app.clickup.com/t/8699zkbej)

antonijzelinskij:
Release severity: Normal
Release notes:
Built Play Integrity API to Protect DApps from Abuse

edogggy (approved):

feature-dapp-impl/src/main/java/io/novafoundation/nova/feature_dapp_impl/utils/integrityCheck/IntegrityCheckSession.kt
```diff
@@ -0,0 +1,121 @@
...
+
+    suspend fun restartIntegrityCheck() {
+        runAttestation()
+
+        runVerifying()
+    }
+
+    fun removeCallback() {
+        callback = null
+    }
```
valentunn, line 61: This looks strange - why do we nullify somthing that has been passed in the constructor? As I see we only use it together with nullifying session itself - i think this solely is enough to enable gc to clean session instance, and thus - callback

valentunn (approved):

PR #2068, opened 2025-08-07
Fixed integrity check pr notes

antonijzelinskij:
Waiting for review to create an internal test build to check everything


common/src/main/java/io/novafoundation/nova/common/utils/IntegrityService.kt
```diff
@@ -5,50 +5,76 @@ import com.google.android.play.core.integrity.StandardIntegrityManager
...
+            state.collect {
+                if (it is ProviderState.Ready) return@collect
+                if (it is ProviderState.ReceivingError) return@collect
+            }
+        }
+
+        if (state.value is ProviderState.Ready) return
+        if (state.value is ProviderState.ReceivingError) {
+            prepareTokenProvider()
+            ensureProviderIsReady(retry + 1)
```
valentunn, line 62: 1. This is replacable with a single `Mutex`
2. I think we can extend `SingleValueCache` to support failable operations (lambda returns `Result<T>`) and leave `SingleValueCache(compute: () -> T)` that modifies `compute` by wrapping `T` to `Result.Success`. This way we can continue using `SingleValueCache` for failable and non-failable operations. And all of this code will just become `val provider = SingleValueCache { prepareTokenProvider() }`

feature-dapp-impl/src/main/java/io/novafoundation/nova/feature_dapp_impl/utils/integrityCheck/IntegrityCheckSession.kt
```diff
@@ -56,14 +56,9 @@ class IntegrityCheckSession(
         runVerifying()
     }
 
-    fun removeCallback() {
-        callback = null
```
valentunn, line 60: `callback` can be `val callback: Callback` then

common/src/main/java/io/novafoundation/nova/common/utils/IntegrityService.kt
antonijzelinskij: I think my solution is good enough since it considers that integrity provider may be in preparing process and doesn't request a new provider in that case just awaits preparing one. So I think using mutex here is not good idea.
We also need to prepare it when app is started (due to google documentation) so I made `prepareTokenProvider()` not `suspend`. This is app-level service and I don't like to use any coroutine scope in `init`

What do you think?

PR #2085, opened 2025-08-15
Fix/utility batch format

antonijzelinskij:
#869a20hcd
image image


stepanLav:
Task linked: [CU-869a20hcd Update Batch formatting on tx list/approve/reject screens](https://app.clickup.com/t/869a20hcd)

feature-multisig/operations/src/main/java/io/novafoundation/nova/feature_multisig_operations/presentation/callFormatting/formatters/RealMultisigCallFormatter.kt
```diff
@@ -148,20 +144,23 @@ class RealMultisigCallFormatter @Inject constructor(
...
+            if (isSingleCallForDepth) {
+                val singleCall = calls.first()
+                val isFirstCallFormatted = singleCall.isFormatted()
+                if (isFirstCallFormatted) return singleCall
+            } else {
+                return@forEach
+            }
         }
+
+        return null
```
antonijzelinskij, line 163: Since now we need to format Utility: Batch (that is in CallTraversal nodes) I found that we need to know call depth.
Hope this logic is also well

PR #2086, opened 2025-08-15
Improve weight heuristics function

valentunn:
* Increase granularity of weights by increasing base weight from 10 to 100
* Increase weight of AssetConversionSwap to 3x cross-chain tranfser weight (see reasoning in the comment)
* Decrease weight of hydration edges when they are merged together into single tx
* Add logging of the weight for better debugging

Main reason for the changes is to fix AssetConversion swaps taking all the candidate spaces not even allowing Hydration paths to be quoted

### DIFF
--- a/feature-swap-core/api/src/main/java/io/novafoundation/nova/feature_swap_core_api/data/paths/model/QuotedPath.kt
+++ b/feature-swap-core/api/src/main/java/io/novafoundation/nova/feature_swap_core_api/data/paths/model/QuotedPath.kt
@@ -24,25 +24,29 @@ class QuotedPath<E>(
     }
 }
 
-class WeightBreakdown(
+class WeightBreakdown private constructor(
     val individualWeights: List<Int>,
     val total: Int
-)
+) {
 
-fun <N, E : WeightedEdge<N>> QuotedPath<E>.weightBreakdown(): WeightBreakdown {
-    val weightedPath = mutableListOf<E>()
-    val individualWeights = mutableListOf<Int>()
-    var weight = 0
+    companion object {
 
-    path.forEach { quotedEdge ->
-        val edgeWeight = quotedEdge.edge.weightForAppendingTo(weightedPath)
+        fun <N, E : WeightedEdge<N>> fromQuotedPath(path: QuotedPath<E>): WeightBreakdown {
+            val weightedPath = mutableListOf<E>()
+            val individualWeights = mutableListOf<Int>()
+            var weight = 0
 
-        weight += edgeWeight
-        weightedPath += quotedEdge.edge
-        individualWeights += edgeWeight
+            path.path.forEach { quotedEdge ->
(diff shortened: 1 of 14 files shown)


feature-swap-core/api/src/main/java/io/novafoundation/nova/feature_swap_core_api/data/paths/model/QuotedPath.kt
```diff
@@ -23,6 +24,27 @@ class QuotedPath<E>(
     }
 }
 
+class WeightBreakdown(
+    val individualWeights: List<Int>,
+    val total: Int
+)
```
antonijzelinskij, line 30: Potentially we may have an invalid state here so I think we can have property `val total = individualWeights.sum()`

feature-swap-core/src/main/java/io/novafoundation/nova/feature_swap_core/data/assetExchange/conversion/types/hydra/sources/Wegiths.kt
```diff
@@ -1,21 +1,34 @@
...
+            }
+        }
+
         const val OMNIPOOL = QuotableEdge.DEFAULT_SEGMENT_WEIGHT
 
-        const val STABLESWAP = QuotableEdge.DEFAULT_SEGMENT_WEIGHT - 1
+        const val STABLESWAP = QuotableEdge.DEFAULT_SEGMENT_WEIGHT - 10
 
-        const val XYK = QuotableEdge.DEFAULT_SEGMENT_WEIGHT + 1
+        const val XYK = QuotableEdge.DEFAULT_SEGMENT_WEIGHT + 10
```
antonijzelinskij, line 23: It seems to me that it is not entirely clear to those who will look at this code for the first time why we divide exactly 10. Can we use Float for weights or at lease left a comment?

antonijzelinskij (approved):

feature-swap-core/src/main/java/io/novafoundation/nova/feature_swap_core/data/assetExchange/conversion/types/hydra/sources/Wegiths.kt
valentunn: Int is just much faster to use than float when it comes to calculations e.t.c
I will leave a comment why it is ten

feature-swap-core/api/src/main/java/io/novafoundation/nova/feature_swap_core_api/data/paths/model/QuotedPath.kt
valentunn: Yeah I just wanted to avoid extra calculation given that the sum is already known at the time `WeightBreakdown` is contrsucted
I am especially carefull with this as it is sort of hot path for us

PR #2094, opened 2025-08-20
Feature/ms pushes notify user

antonijzelinskij:
#869a20ehb
image


feature-account-impl/src/main/java/io/novafoundation/nova/feature_account_impl/data/proxy/RealMetaAccountsUpdatesRegistry.kt
```diff
@@ -57,6 +60,13 @@ class RealMetaAccountsUpdatesRegistry(
             .map { hasUpdates() }
     }
 
+    override fun observeConsumedUpdatesMetaIds(): Flow<Set<Long>> {
```
valentunn, line 63: `observeLastConsumedUpdatesMetaIds`
Otherwise it reads like we are accumulating all previosuly consumed ids

feature-push-notifications/src/main/java/io/novafoundation/nova/feature_push_notifications/domain/interactor/MultisigPushAlertInteractor.kt
```diff
@@ -0,0 +1,69 @@
...
+
+    override fun setAlertWasAlreadyShown() {
+        multisigPushAlertRepository.setMultisigsPushAlertWasShown()
+    }
+
+    /**
+     * We have to check if we can show alert right after user update the app.
+     * Showing is allowed when user have multisig accounts in first app start after update
+     */
+    override suspend fun allowedToShowAlertAtStart(): Boolean {
```
valentunn, line 48: This function has implicit side-effect - as a client when I read `allowedToShowAlertAtStart` I just expect function to calculate me the value, and not update some setting e.t.c

feature-push-notifications/src/main/java/io/novafoundation/nova/feature_push_notifications/domain/interactor/MultisigPushAlertInteractor.kt
```diff
@@ -0,0 +1,69 @@
...
+        val allowingState = multisigPushAlertRepository.showAlertAtStartAllowingState()
+
+        if (allowingState == AllowingState.INITIAL) {
+            val userHasMultisigs = accountRepository.hasMetaAccountsByType(LightMetaAccount.Type.MULTISIG)
+            if (userHasMultisigs) {
+                multisigPushAlertRepository.setAlertAtStartAllowingState(AllowingState.ALLOWED)
+                return true
+            } else {
+                multisigPushAlertRepository.setAlertAtStartAllowingState(AllowingState.NOT_ALLOWED)
+                return false
```
valentunn, line 58: Why do we even need to store the state? You check for the presence of meta accounts anyway, you can just return this boolean. 
Also there is a bug - once we set the value to NOT_ALLOWED (becuase there are no msigs), the value can never be chagned because setAlertAtStartAllowingState is only called when state is INITIAL

P.S. it seems like we can just return `!multisigPushAlertRepository.isAlertAlreadyShown() && userHasMultisigs`

feature-push-notifications/src/main/java/io/novafoundation/nova/feature_push_notifications/presentation/multisigsWarning/MultisigPushNotificationsAlertMixin.kt
```diff
@@ -0,0 +1,68 @@
...
+}
+
+class RealMultisigPushNotificationsAlertMixin(
+    private val automaticInteractionGate: AutomaticInteractionGate,
+    private val interactor: MultisigPushAlertInteractor,
+    private val metaAccountsUpdatesRegistry: MetaAccountsUpdatesRegistry,
+    private val router: PushNotificationsRouter
+) : MultisigPushNotificationsAlertMixin {
+
+    override val showWarningEvent = MutableLiveData<Event<Unit>>()
```
valentunn, line 33: Why is it a warning?

feature-push-notifications/src/main/java/io/novafoundation/nova/feature_push_notifications/domain/interactor/MultisigPushAlertInteractor.kt
```diff
@@ -0,0 +1,69 @@
...
+import io.novafoundation.nova.feature_account_api.domain.model.LightMetaAccount
+import io.novafoundation.nova.feature_account_api.domain.model.isMultisig
+import io.novafoundation.nova.feature_push_notifications.data.repository.MultisigPushAlertRepository
+import io.novafoundation.nova.feature_push_notifications.data.settings.PushSettingsProvider
+
+interface MultisigPushAlertInteractor {
+
+    fun isPushNotificationsEnabled(): Boolean
+
+    fun isAlertWasAlreadyShown(): Boolean
```
valentunn, line 13: `wasAlertAlreadyShown` or `isAlertAlreadyShown`

feature-push-notifications/src/main/java/io/novafoundation/nova/feature_push_notifications/presentation/multisigsWarning/MultisigPushNotificationsAlertMixin.kt
```diff
@@ -0,0 +1,68 @@
...
+        if (interactor.isAlertWasAlreadyShown()) return@launchUnit
+
+        // We should get this state before multisigs will be discovered so we call this method before interaction gate
+        val allowedToShowAlertAtStart = interactor.allowedToShowAlertAtStart()
+
+        automaticInteractionGate.awaitInteractionAllowed()
+
+        if (allowedToShowAlertAtStart) {
+            interactor.setAlertWasAlreadyShown()
+            showWarningEvent.sendEvent()
```
valentunn, line 45: This can be extracted into a separate function since we do it twice

valentunn (approved):

feature-push-notifications/src/main/java/io/novafoundation/nova/feature_push_notifications/domain/interactor/MultisigPushAlertInteractor.kt
antonijzelinskij: Yes but I think naming is good because that's what is happening inside - we try to understand if we can show alert at start. I'll explain how it works in another comment

feature-push-notifications/src/main/java/io/novafoundation/nova/feature_push_notifications/domain/interactor/MultisigPushAlertInteractor.kt
antonijzelinskij: We need to store this state to achieve that behaviour:
-  If user have multisigs and update the app - we will show him this bottom sheet right after pin was entered
- If user doesn't have multisigs and update the app and after that add multisig account - we will show bottom sheet after multisig-wallet-updates will be consumed. And since user may close app before consuming wallet-updates and start it again we need to have some state to check if user has multisigs or not right after app was updated.

If you have any recommendations how to make this code more clear please let me know

PR #2096, opened 2025-08-20
Fixed sync push notification settings issues

antonijzelinskij:
#869a20ehf
- consider affected accounts when remove meta account
- fixed issue when push settings was not syncing at all

### DIFF
--- a/feature-account-impl/src/main/java/io/novafoundation/nova/feature_account_impl/data/repository/AccountRepositoryImpl.kt
+++ b/feature-account-impl/src/main/java/io/novafoundation/nova/feature_account_impl/data/repository/AccountRepositoryImpl.kt
@@ -15,6 +15,7 @@ import io.novafoundation.nova.core_db.model.AccountLocal
 import io.novafoundation.nova.core_db.model.NodeLocal
 import io.novafoundation.nova.feature_account_api.data.events.MetaAccountChangesEventBus
 import io.novafoundation.nova.feature_account_api.data.events.MetaAccountChangesEventBus.Event
+import io.novafoundation.nova.feature_account_api.data.events.combineBusEvents
 import io.novafoundation.nova.feature_account_api.data.secrets.keypair
 import io.novafoundation.nova.feature_account_api.domain.interfaces.AccountRepository
 import io.novafoundation.nova.feature_account_api.domain.model.Account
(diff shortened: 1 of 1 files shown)


stepanLav:
Task linked: [CU-869a20ehf Add sending a request to deactivate notifications to the process of removing a wallet from Nova](https://app.clickup.com/t/869a20ehf)

core-db/src/main/java/io/novafoundation/nova/core_db/dao/MetaAccountDao.kt
```diff
@@ -188,17 +188,29 @@ interface MetaAccountDao {
...
+        JOIN accounts_to_delete r ON m.parentMetaId = r.id
+    )
+    SELECT id FROM accounts_to_delete
     """
     )
-    suspend fun delete(metaId: Long)
+    suspend fun findIdsToDelete(metaId: Long): List<Long>
+
+    @Query("DELETE FROM meta_accounts WHERE id IN (:ids)")
+    suspend fun deleteByIds(ids: List<Long>)
```
valentunn, line 204: JFYI I found a solution to do that in a single SQL query (via `RETURNING` keyword) but its only available in sqlite build that is included in Android 11+

feature-account-impl/src/main/java/io/novafoundation/nova/feature_account_impl/data/repository/AccountRepositoryImpl.kt
```diff
@@ -173,9 +173,10 @@ class AccountRepositoryImpl(
     override suspend fun deleteAccount(metaId: Long) = withContext(Dispatchers.Default) {
         val metaAccountType = accountDataSource.getMetaAccountTypeOrThrow(metaId)
 
-        accountDataSource.deleteMetaAccount(metaId)
+        val allAffectedMetaIds = accountDataSource.deleteMetaAccount(metaId)
 
-        metaAccountChangesEventBus.notify(Event.AccountRemoved(metaId, metaAccountType), source = null)
+        val deleteEvents = allAffectedMetaIds.map { Event.AccountRemoved(it, metaAccountType) }
+        metaAccountChangesEventBus.notify(Event.BatchUpdate(deleteEvents), source = null)
```
valentunn, line 179: There is an extension `combineBusEvents to convert `List< Event>` to a batch. It also takes into account that there is no need to wrap into a batch if there is just a single event

PR #2105, opened 2025-08-28
Update multisig push settings on meta account delete

antonijzelinskij:
#869a8aefu
#869a8fmwm

### DIFF
--- a/feature-push-notifications/src/main/java/io/novafoundation/nova/feature_push_notifications/domain/interactor/PushNotificationsInteractor.kt
+++ b/feature-push-notifications/src/main/java/io/novafoundation/nova/feature_push_notifications/domain/interactor/PushNotificationsInteractor.kt
@@ -1,6 +1,7 @@
 package io.novafoundation.nova.feature_push_notifications.domain.interactor
 
 import io.novafoundation.nova.feature_account_api.domain.interfaces.AccountRepository
+import io.novafoundation.nova.feature_account_api.domain.model.LightMetaAccount
 import io.novafoundation.nova.feature_account_api.domain.model.MetaAccount
 import io.novafoundation.nova.feature_push_notifications.data.PushNotificationsAvailabilityState
 import io.novafoundation.nova.feature_push_notifications.data.PushNotificationsService
@@ -37,7 +38,9 @@ interface PushNotificationsInteractor {
 
     suspend fun onMetaAccountChange(changed: List<Long>, deleted: List<Long>)
 
-    suspend fun filterUnavailableMetaIds(metaIds: Set<Long>): Set<Long>
+    suspend fun filterAvailableMetaIdsAndGetNewState(pushSettings: PushSettings): PushSettings
+
+    suspend fun getNewStateForChangedMetaAccounts(currentSettings: PushSettings, newMetaIds: Set<Long>): PushSettings
 }
 
 class RealPushNotificationsInteractor(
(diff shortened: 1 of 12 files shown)


feature-push-notifications/src/main/java/io/novafoundation/nova/feature_push_notifications/domain/interactor/PushNotificationsInteractor.kt
```diff
@@ -111,14 +114,30 @@ class RealPushNotificationsInteractor(
...
-                pushNotificationsService.updatePushSettings(enabled = newEnabledState, pushSettings = newPushSettingsOrNull)
+                val newPushSettings = getNewStateForChangedMetaAccounts(pushSettings, newSubscribedMetaAccounts)
+                pushNotificationsService.updatePushSettings(enabled = newEnabledState, pushSettings = newPushSettings)
             }
         }
     }
 
-    override suspend fun filterUnavailableMetaIds(metaIds: Set<Long>): Set<Long> {
-        return accountRepository.getUnavailableMetaIdsFromSet(metaIds)
+    override suspend fun filterAvailableMetaIdsAndGetNewState(pushSettings: PushSettings): PushSettings {
```
valentunn, line 123: Maybe `refreshStateBasedOnAvailableAccounts`?

feature-push-notifications/src/main/java/io/novafoundation/nova/feature_push_notifications/domain/interactor/PushNotificationsInteractor.kt
```diff
@@ -111,14 +114,30 @@ class RealPushNotificationsInteractor(
...
+        val availableMetaIds = accountRepository.getAvailableMetaIdsFromSet(pushSettings.subscribedMetaAccounts)
+        if (availableMetaIds == pushSettings.subscribedMetaAccounts) return pushSettings
+
+        return getNewStateForChangedMetaAccounts(pushSettings, availableMetaIds)
+    }
+
+    override suspend fun getNewStateForChangedMetaAccounts(currentSettings: PushSettings, newMetaIds: Set<Long>): PushSettings {
```
valentunn, line 130: I think there is one more place we should fix - when we react on account deletion in `PushSettingsSyncRequestBusHandler` - seems like we are not updating msig settings in onMetaAccountChange
Also a few notes here:

1. MetaAccountChangesEventBus.Event already contain MetaAccount type alongside its id
2. BUT we specify this type wrongly in AccountRepositoryImpl.deleteAccount - we use account type of initially deleted account for each of deleted dependents. This might break other features of the app in unpredictable way so I suggest we fix this while we are working on this topic

feature-push-notifications/src/main/java/io/novafoundation/nova/feature_push_notifications/domain/interactor/PushNotificationsInteractor.kt
antonijzelinskij: Refresh sounds like an action that change something in data source

feature-push-notifications/src/main/java/io/novafoundation/nova/feature_push_notifications/domain/interactor/PushNotificationsInteractor.kt
antonijzelinskij: Fixed

valentunn (approved):

PR #2107, opened 2025-09-01
Fixed account duplication

antonijzelinskij:
#869a6wcdc
Fixed accounts update notifications for deactivated meta accounts

### DIFF
--- a/feature-account-impl/src/main/java/io/novafoundation/nova/feature_account_impl/data/sync/RealExternalAccountsSyncService.kt
+++ b/feature-account-impl/src/main/java/io/novafoundation/nova/feature_account_impl/data/sync/RealExternalAccountsSyncService.kt
@@ -3,12 +3,13 @@ package io.novafoundation.nova.feature_account_impl.data.sync
 import android.util.Log
 import io.novafoundation.nova.common.address.AccountIdKey
 import io.novafoundation.nova.common.di.scope.FeatureScope
+import io.novafoundation.nova.common.utils.associateMutableBy
 import io.novafoundation.nova.common.utils.coroutines.RootScope
 import io.novafoundation.nova.common.utils.filterToSet
 import io.novafoundation.nova.common.utils.launchUnit
 import io.novafoundation.nova.common.utils.mapToSet
-import io.novafoundation.nova.common.utils.mutableMultiListMapOf
 import io.novafoundation.nova.common.utils.put
+import io.novafoundation.nova.common.utils.toMutableMultiMapList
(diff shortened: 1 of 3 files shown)


stepanLav:
Task linked: [CU-869a6wcdc Android. Duplicated ms wallets added](https://app.clickup.com/t/869a6wcdc)

feature-account-impl/src/main/java/io/novafoundation/nova/feature_account_impl/data/sync/RealExternalAccountsSyncService.kt
```diff
@@ -300,7 +298,7 @@ internal class RealExternalAccountsSyncService @Inject constructor(
                     return@controllersLoop
                 }
 
-                val existingControlledAccounts = existingAccountsByParentId[controller.id].orEmpty()
+                val existingControlledAccounts = controllersByParentId[controller.id].orEmpty()
```
valentunn, line 301: Can you please describe a case when this change matters? Am I right that this fixes dublication in case some multisig is present twice in the list and when we process is second time, we dont see the account added in the first iteration since `existingAccountsByParentId` only contain snapshot of the db accounts before the processing start?

If so, can you please describe the case itself, like what is the structure of accounts e.t.c. that causes the same account to be present multiple times in the `foundExternalAccounts`

valentunn (approved):

feature-account-impl/src/main/java/io/novafoundation/nova/feature_account_impl/data/sync/RealExternalAccountsSyncService.kt
```diff
@@ -138,12 +139,13 @@ internal class RealExternalAccountsSyncService @Inject constructor(
...
 
-    private suspend fun notifyAboutAddedAccounts(added: List<AddAccountResult.AccountAdded>) {
+    private suspend fun notifyAboutAddedAccounts(added: List<AddAccountResult.AccountAdded>, deactivatedMetaIds: Set<Long>) {
         added.map { it.toAccountBusEvent() }
             .combineBusEvents()
             ?.let { eventBus.notify(it, source = ACCOUNTS_CHANGED_SOURCE) }
 
-        metaAccountsUpdatesRegistry.addMetaIds(added.map { it.metaId })
+        val updatedAccountsIds = deactivatedMetaIds.toList() + added.map { it.metaId }
+        metaAccountsUpdatesRegistry.addMetaIds(updatedAccountsIds)
```
valentunn, line 148: On a separate note: I think we should improve the naming of methods in this class since I thought we only need to add `added` updates there

PR #2109, opened 2025-09-04
Fix/governance delegations timepoint

antonijzelinskij:
#8699qvjag


stepanLav:
Task linked: [CU-8699qvjag AH Governance - Add timestamp from Indexers support for "30 days vote count" case](https://app.clickup.com/t/8699qvjag)

PR #2132, opened 2025-09-26
Chain migration details screen + deep links

antonijzelinskij:
#869ajyuvc
image


feature-ahm-impl/src/main/java/io/novafoundation/nova/feature_ahm_impl/data/repository/RealChainMigrationRepository.kt
```diff
@@ -0,0 +1,75 @@
...
+    private suspend fun getLastBlockNumber(chainId: String): BigInteger {
+        return remoteStorageSource.query(chainId) {
+            runtime.metadata.system().storage("Number").query {
+                it.cast<BigInteger>()
+            }
+        }
+    }
+
+    private fun saveToStorage(chainsWithAssetBalance: Set<String>) {
+        val currentChains = getChainsWithAssetBalance()
```
valentunn, line 66: Why do we need to cache them to pestistent storage? Looks like saving them in-memory (e.g. in the member property of `RealChainMigrationRepository`

feature-ahm-impl/src/main/java/io/novafoundation/nova/feature_ahm_impl/data/repository/RealMigrationInfoRepository.kt
```diff
@@ -0,0 +1,40 @@
...
+import io.novafoundation.nova.feature_ahm_impl.data.config.toDomain
+import kotlinx.coroutines.sync.Mutex
+import kotlinx.coroutines.sync.withLock
+
+class RealMigrationInfoRepository(
+    private val api: ChainMigrationConfigApi
+) : MigrationInfoRepository {
+
+    private var configs: List<ChainMigrationConfig>? = null
+    private val mutex = Mutex()
```
valentunn, line 15: Looks like we can use `SingleValueCache` here

feature-ahm-impl/src/main/java/io/novafoundation/nova/feature_ahm_impl/presentation/migrationDetails/ChainMigrationDetailsViewModel.kt
```diff
@@ -0,0 +1,100 @@
...
+import io.novafoundation.nova.feature_banners_api.presentation.source.BannersSourceFactory
+import io.novafoundation.nova.feature_banners_api.presentation.source.forDirectory
+import java.text.SimpleDateFormat
+import java.util.Locale
+import kotlinx.coroutines.flow.combine
+import kotlinx.coroutines.flow.first
+import kotlinx.coroutines.flow.map
+import kotlinx.coroutines.launch
+
+class ConfigModel(
```
valentunn, line 25: I suggest to extract it to a separate file

valentunn (approved):

feature-ahm-impl/src/main/java/io/novafoundation/nova/feature_ahm_impl/data/repository/RealChainMigrationRepository.kt
antonijzelinskij: I wanted to prevent a case when user open app and close it on pin entering.
In this case we may already sync balances and in next opening migration details wouldn't be shown

PR #2140, opened 2025-10-01
Fix paged exposures sync

valentunn:
* Force using paged exposures when we see that legacy exposures are no longer available in runtime
* Refactor logic to be more understandable and concise
* Remove runCatching in fetching so the exception will stop the updater but wont result in wrong data being written to the flag storage

### DIFF
--- a/feature-staking-impl/src/main/java/io/novafoundation/nova/feature_staking_impl/data/network/blockhain/updaters/ValidatorExposureUpdater.kt
+++ b/feature-staking-impl/src/main/java/io/novafoundation/nova/feature_staking_impl/data/network/blockhain/updaters/ValidatorExposureUpdater.kt
@@ -10,10 +10,10 @@ import io.novafoundation.nova.core.updater.Updater
 import io.novafoundation.nova.feature_staking_api.domain.model.EraIndex
 import io.novafoundation.nova.feature_staking_impl.data.StakingSharedState
 import io.novafoundation.nova.feature_staking_impl.data.network.blockhain.updaters.ValidatorExposureUpdater.Companion.STORAGE_KEY_PAGED_EXPOSURES
-import io.novafoundation.nova.runtime.network.updaters.multiChain.SharedStateBasedUpdater
 import io.novafoundation.nova.feature_staking_impl.data.network.blockhain.updaters.scope.ActiveEraScope
 import io.novafoundation.nova.runtime.multiNetwork.ChainRegistry
 import io.novafoundation.nova.runtime.multiNetwork.getRuntime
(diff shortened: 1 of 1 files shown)


feature-staking-impl/src/main/java/io/novafoundation/nova/feature_staking_impl/data/network/blockhain/updaters/ValidatorExposureUpdater.kt
```diff
@@ -126,29 +126,60 @@ class ValidatorExposureUpdater(
...
+    }
+
+    private fun RuntimeSnapshot.detectExposureStateFromPallets(): ExposureState {
+        return when {
+            legacyExposuresFullyRemoved() -> ExposureState.CERTAIN_PAGED
+            // Just because paged exposures are enabled does not mean they are actually used yet
+            pagedExposuresEnabled() -> ExposureState.UNCERTAIN
+            else -> ExposureState.CERTAIN_LEGACY
+        }
+    }
```
antonijzelinskij, line 157: I don't understand clearly why if we can't use legacy exposures in case when we have legacy and pages in the same time?
Could you please explain it to me?

antonijzelinskij (approved):

feature-staking-impl/src/main/java/io/novafoundation/nova/feature_staking_impl/data/network/blockhain/updaters/ValidatorExposureUpdater.kt
valentunn: The thing here is that we have both **storages** in runtime. However, it for a given era, only one of them is filled.
Overall, the legacy - paged transition flow is the following:

1. Initial. legacy: present in runtime, has value; paged: not present in runtime
2. Paged exposures are added.   legacy: present in runtime, has value; paged: present in runtime
3. New era after (2) comes.  legacy: present in runtime, has value for older eras still. paged: present in runtime. has value for new era
4. <HistoryDepth> number of eras passes.  legacy: present in runtime, fully empty for all eras; present in runtime. has value for all  <HistoryDepth> eras
5. In the follow up upgrade, legacy exposure storage is removed. legacy: not present in runtime paged:  present in runtime, has value;

A special note should be given in the context of AHM:
During AHM, paged exposures, as other data, will be moved from Kusama to KAH, so during the migration we might be actually in quite an inconsistent state. This is only partially addressed by giving preference to pallet presence. In particular, returning CERTAIN_PAGED when we detect that legacy exposures are not in runtime helps to avoid the crash when repository tries to read legacy exposures

PR #2179, opened 2025-10-27
Select gift amount

antonijzelinskij:
#869av2n8v
image image image image

### DIFF
--- a/feature-gift-impl/src/main/java/io/novafoundation/nova/feature_gift_impl/presentation/amount/GiftMinAmountProvider.kt
+++ b/feature-gift-impl/src/main/java/io/novafoundation/nova/feature_gift_impl/presentation/amount/GiftMinAmountProvider.kt
@@ -10,6 +10,7 @@ import kotlinx.coroutines.flow.Flow
 import kotlinx.coroutines.flow.combine
 import kotlinx.coroutines.flow.filterNotNull
 import kotlinx.coroutines.flow.map
+import kotlinx.coroutines.flow.mapNotNull
 
 class GiftMinAmountProviderFactory(
     private val createGiftInteractor: CreateGiftInteractor,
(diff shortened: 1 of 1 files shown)


stepanLav:
Task linked: [CU-869av2n8v Add amount input screen](https://app.clickup.com/t/869av2n8v)

feature-gift-impl/src/main/java/io/novafoundation/nova/feature_gift_impl/domain/CreateGiftInteractor.kt
```diff
@@ -0,0 +1,87 @@
...
+            transferMax = true,
+            coroutineScope = coroutineScope
+        )
+        val claimFeeAmount = model.chainAsset.amountFromPlanks(claimGiftFee.amount)
+
+        val createGiftFee = getSubmissionFee(
+            model = model.copy(amount = model.amount + claimFeeAmount),
+            transferMax = transferAllToCreateGift,
+            coroutineScope = coroutineScope
+        )
```
valentunn: We can save some loading time here by only using 2 * createGiftFee. This is safe since fee for the initial transfer is greater or equal to the fee of the follow up 

Worst case we can still load fee faster by loading `claimGiftFee` and `createGiftFee` concurrently

feature-gift-impl/src/main/java/io/novafoundation/nova/feature_gift_impl/presentation/amount/GiftMinAmountProvider.kt
```diff
@@ -0,0 +1,49 @@
...
+        val chainAssetWithEDFlow = feeMixin.feeChainAssetFlow
+            .map { it to createGiftInteractor.getExistentialDeposit(it) }
+
+        return combine(
+            feeFlow,
+            chainAssetWithEDFlow
+        ) { fee, chainAssetWithEd ->
+            val chainAsset = chainAssetWithEd.first
+            val existentialDeposit = chainAssetWithEd.second
+            chainAsset.amountFromPlanks(fee.amount) + existentialDeposit
```
valentunn: We only need to limit by `fee.claimGiftFee` actually. Not by `claimGiftFee + createGiftFee`

valentunn (approved):

PR #2212, opened 2025-12-03
NW-1687 Scan QR seed

antonijzelinskij:
image image

### DIFF
--- a/feature-account-impl/src/main/java/io/novafoundation/nova/feature_account_impl/domain/paritySigner/connect/finish/FinishImportParitySignerInteractor.kt
+++ b/feature-account-impl/src/main/java/io/novafoundation/nova/feature_account_impl/domain/paritySigner/connect/finish/FinishImportParitySignerInteractor.kt
@@ -5,7 +5,6 @@ import io.novafoundation.nova.feature_account_api.data.repository.addAccount.Add
 import io.novafoundation.nova.feature_account_api.data.repository.addAccount.addAccountWithSingleChange
 import io.novafoundation.nova.feature_account_api.domain.account.advancedEncryption.AdvancedEncryption
 import io.novafoundation.nova.feature_account_api.domain.account.advancedEncryption.substrate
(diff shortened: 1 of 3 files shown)


feature-account-impl/src/main/java/io/novafoundation/nova/feature_account_impl/presentation/importing/source/source/RawSeedImportSource.kt
```diff
@@ -34,17 +37,30 @@ class RawSeedImportSource(
 
     override val fieldsValidFlow: Flow<Boolean> = rawSeedFlow.map { it.isNotEmpty() }
 
+    init {
+        scanSeedRequester.responseFlow
+            .onEach {
+                // A short delay to wait until view will be attached to fragment. Otherwise the input view clear provided value on attach to fragment
+                delay(150)
```
valentunn: We should probably just remove saving instnace state for input field. We did the same for SwapSettingsFragment when dealing with input fields after selecting a swap asset

feature-account-impl/src/main/java/io/novafoundation/nova/feature_account_impl/domain/scanSeed/ScanSeedInteractor.kt
```diff
@@ -0,0 +1,23 @@
...
+
+class RealScanSeedInteractor(
+    private val secretQrFormat: SecretQrFormat
+) : ScanSeedInteractor {
+
+    override fun decodeSeed(content: String): Result<String> = runCatching {
+        val payload = secretQrFormat.decode(content)
+
+        when (val secret = payload.secret) {
+            is ScanSecret.EncryptedKeypair -> error("Encoded key can't be decoded to seed")
```
valentunn: Is this a confirmed decision? Tbh I see no issue with handling that. We just need to adjust the `RawSeedImportSource` and its interactor to support that

valentunn (approved):

PR #2268, opened 2026-03-20
Force Nova pool for Asset Hub and lock Novasama validators in selection

Leemo94:
- Bypass pool selection screen for Polkadot/Kusama Asset Hub chains; recommendation system selects Nova pool and user cannot change it
- Pool name hidden on staking type screen; "Recommended" label persists
- Preferred (Nova) validators locked at bottom of custom validator list with greyed-out checkboxes; cannot be deselected via tap, search, edit-mode delete, or "Deselect All"
- Selection count reflects community-only available slots

### DIFF
--- a/feature-staking-impl/src/main/java/io/novafoundation/nova/feature_staking_impl/data/nominationPools/pool/KnownNovaPools.kt
+++ b/feature-staking-impl/src/main/java/io/novafoundation/nova/feature_staking_impl/data/nominationPools/pool/KnownNovaPools.kt
@@ -1,9 +1,15 @@
 package io.novafoundation.nova.feature_staking_impl.data.nominationPools.pool
 
 import io.novafoundation.nova.feature_staking_api.domain.nominationPool.model.PoolId
+import io.novafoundation.nova.runtime.ext.ChainGeneses
 import io.novafoundation.nova.runtime.ext.Geneses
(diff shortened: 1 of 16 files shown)


Leemo94:
CI failures here seem to be the standard fork-secrets restriction (scaleway/action-scw-secret can't authenticate without the access key, which GitHub doesn't expose to fork PRs).

valentunn (approved):

Leemo94:
1. User enters nomination pool
Screenshot_1778844508
Screenshot_1778844510

This update disables the ability to select "pool staking" to change nomination pools. There are examples where >10% of all nomination pool stakers (who arent in a wallet-operators pool) get scammed by the third-parity nom pool operator raising commission, etc. This change stops users from falling victim to this. I can share more information about this investigation if needed.


2. Sticky Novasama Validators
Screenshot_1778844514
Screenshot_1778844517
Screenshot_1778844519
Screenshot_1778844523

The flow here remains the same, except the checkbox next to Novasama affiliated validators (as per the list) is greyed out and cannot be deselected. We are able to control the state of our validators to ensure they meet the network minimums and are in the active set / do not raise commission on users. We are not able to control other operators behaviour. Additionally, the state of funding in the ecosystem leads me to believe that this is something we should do to increase revenue (regardless of how low rewards are).

PR #2269, opened 2026-03-20
DApp browser — staking navigation improvements

Leemo94:
- Show informational banner in DApp search when staking-related queries are detected, guiding users to native staking
- Display notice when navigating to third-party staking sites, with option to continue or go to built-in staking
- Add staking keyword detection for search queries (Latin, Cyrillic, CJK)
- Add domain-level detection for known third-party staking sites


Leemo94:
### 1. Staking banner in Dapp catalog search
Screenshot_1778778727

When a user enters staking related keywords in the dapp catalog, they are presented with a button to go to Nova Wallet's native staking.

### 2. Display Notice when navigating to third party staking sites


Screenshot_1778778734


Screenshot_1778778737

Due to the current funding situation in Polkadot & user comments in our chats & support email – this PR adds a notice that the user can manage their DOT stake directly using Nova Wallet, and not using third-party sites which may become deprecated. Some users believe they are "staking with" a particular application, which is not true.



### 3. Localizations/Translations

Screenshot_1778778791


Screenshot_1778778818

Dapp catalog search should support keywords in different languages.

app/src/main/java/io/novafoundation/nova/app/root/navigation/navigators/dApp/DAppNavigator.kt
```diff
@@ -76,6 +78,13 @@ class DAppNavigator(
             .navigateInFirstAttachedContext()
     }
 
+    override fun navigateToStaking() {
+        // DApp browser/search lives in the root nav graph, so we pop the root back stack
+        // to return to the split screen, then switch to the staking tab via bottom navigation.
+        back()
+        stakingDashboardNavigator.openStakingDashboard()
```
antonijzelinskij: Not a completely safe solution - we can open search right in browser clicking on address field. If in this case we call back() we will return to browser - not to main screen so `openStakingDashboard` must not work in this case.

First of all we have to be sure we returned to main screen.
The most easiest way may be this but we also have to check it's work:
```
navigationBuilder().action(R.id.action_open_split_screen).navigateInRoot()
stakingDashboardRouter.openStakingDashboard()
```

app/src/main/java/io/novafoundation/nova/app/root/navigation/navigators/staking/StakingDashboardNavigator.kt
antonijzelinskij, line 1: We don't have to change this file in my openion.
We already had implementation for delayed navigation and we already supported navigation using bottom navigation view.

So all we have to do just be sure we have returned to main screen and call `openStakingDashboard`

So this solution from previous comment is still relevant for us here:
```
navigationBuilder().action(R.id.action_open_split_screen).navigateInRoot()
stakingDashboardRouter.openStakingDashboard()
```

app/src/main/java/io/novafoundation/nova/app/root/presentation/main/MainFragment.kt
```diff
@@ -51,6 +51,7 @@ class MainFragment : BaseFragment<MainViewModel, FragmentMainBinding>() {
 
         binder.bottomNavigationView.setupWithNavController(navController!!)
         binder.bottomNavigationView.itemIconTintList = null
+        stakingDashboardNavigator.setBottomNavigationView(binder.bottomNavigationView)
```
antonijzelinskij: May be removed. We already attached nav controller for bottom navigation view and can use it in StakingDashboardNavigator

common/src/main/res/values-es/strings.xml
```diff
@@ -2039,4 +2039,11 @@
     <string name="yield_boost_stake_increase_time">Tiempo de aumento de stake</string>
     <string name="yield_boost_terms" formatted="false">Yield Boost apostará automáticamente %s todos mis tokens transferibles por encima de %s</string>
     <string name="yiled_boost_yield_boosted">Con Yield Boost</string>
+    <string name="dapp_search_staking_banner_text">¿Quieres hacer staking? Nova Wallet tiene staking integrado.</string>
+    <string name="dapp_search_staking_banner_button">Ir a Staking</string>
+    <string name="dapp_staking_competitor_warning_title">Aviso</string>
+    <string name="dapp_staking_competitor_warning_subtitle">Los sitios de staking de terceros suelen quedar obsoletos. Recomendamos usar las funciones de staking integradas de Nova Wallet.</string>
+    <string name="dapp_staking_competitor_go_to_stake">Ir a Staking</string>
+    <string name="dapp_staking_competitor_advanced">Avanzado</string>
+    <string name="dapp_staking_competitor_continue">Continuar al sitio</string>
```
antonijzelinskij: We translate all string resources using localization service so please remove all strings besides english

feature-dapp-impl/src/main/java/io/novafoundation/nova/feature_dapp_impl/domain/search/StakingKeywordMatcher.kt
```diff
@@ -0,0 +1,103 @@
...
+        // Turkish
+        "stake yapmak", "staking yapmak",
+        "\u00f6d\u00fcller",
+        "do\u011frulay\u0131c\u0131",
+        "delege",
+        // Polish
+        "stakowanie",
+        "nagrody",
+        "walidator", "nominacja",
+        "delegowa\u0107",
```
antonijzelinskij: Please check MULTILINGUAL_KEYWORDS don't have unicode hex symbols like that: \u0441
We can just write whole words in unicode here

feature-dapp-impl/src/main/java/io/novafoundation/nova/feature_dapp_impl/domain/search/StakingKeywordMatcher.kt
```diff
@@ -0,0 +1,103 @@
...
+        "x\u00e1c th\u1ef1c",
+        "\u1ee7y quy\u1ec1n",
+        // Italian
+        "fare staking", "ricompense", "validatore", "nominare", "delegare",
+        // Portuguese
+        "fazer staking", "recompensas", "validador", "nomear", "delegar"
+    )
+
+    // CJK keywords — matched by contains (substring matching)
+    private val CJK_KEYWORDS = listOf(
```
antonijzelinskij, line 67: The same here

feature-dapp-impl/src/main/java/io/novafoundation/nova/feature_dapp_impl/domain/browser/StakingCompetitorDomains.kt
```diff
@@ -0,0 +1,31 @@
...
+        "staking.polkadot.network"
+    )
+
+    fun isStakingCompetitor(url: String): Boolean {
+        return runCatching {
+            val host = Urls.hostOf(url)
+            BLOCKED_DOMAINS.any { domain -> host == domain || host.endsWith(".$domain") }
+        }.getOrDefault(false)
+    }
+}
```
antonijzelinskij: I think it's would be better to keep this sites in nova-utils. So it let us to change it without updating the app and also have only one edit point to keep it the same with iOS.

I suggest to:
- do a pr to nova-utils with this domains
- add an Api to android and iOS to load this domain list
- add a competitors domains provider that will call api.load to upload domains in first call and store it in memory (to don't tall it each time)

feature-dapp-impl/src/main/java/io/novafoundation/nova/feature_dapp_impl/presentation/browser/main/DAppBrowserFragment.kt
```diff
@@ -125,6 +127,15 @@ class DAppBrowserFragment : BaseFragment<DAppBrowserViewModel, FragmentDappBrows
         binder.dappBrowserFavorite.setOnClickListener { viewModel.onFavoriteClick() }
         binder.dappBrowserMore.setOnClickListener { moreClicked() }
 
+        binder.dappBrowserStakingWarningGoToStake.setOnClickListener { viewModel.navigateToStaking() }
+        binder.dappBrowserStakingWarningAdvanced.setOnClickListener {
+            binder.dappBrowserStakingWarningAdvanced.isVisible = false
+            binder.dappBrowserStakingWarningContinue.isVisible = true
+        }
```
antonijzelinskij, line 132: We manage all UI screen state using viewModel that store this state. So firstly view model has to change it's state that will be provided to this screen in flow

feature-dapp-impl/src/main/java/io/novafoundation/nova/feature_dapp_impl/presentation/browser/main/DAppBrowserFragment.kt
```diff
@@ -314,6 +335,14 @@ class DAppBrowserFragment : BaseFragment<DAppBrowserViewModel, FragmentDappBrows
     }
 
     override fun shouldOverrideUrlLoading(view: WebView, request: WebResourceRequest): Boolean {
+        val url = request.url.toString()
+
+        // Check staking competitor domains before the page loads.
+        // onStakingCompetitorIntercepted returns false if the domain has been bypassed by the user.
+        if (StakingCompetitorDomains.isStakingCompetitor(url) && viewModel.onStakingCompetitorIntercepted(url)) {
```
antonijzelinskij: All logic should be on viewModel, so this should be looking like that:
`if (viewModel.onStakingCompetitorIntercepted(url))`
So checking that this url is a competitor must be on view model side.

I also have to notice that using global funs like this is a bad practice. Hope we will use nova-utils and competitor domains provider but if using nova utils is looking like an overcoding let's anyway create a normal provider with hardcoded domains.

feature-dapp-impl/src/main/res/layout/fragment_dapp_browser.xml
```diff
@@ -59,6 +59,80 @@
...
+        <View
+            android:layout_width="0dp"
+            android:layout_height="0dp"
+            android:layout_weight="1" />
+
+        <androidx.appcompat.widget.AppCompatImageView
+            android:layout_width="88dp"
+            android:layout_height="88dp"
+            android:src="@drawable/ic_info_24"
+            tools:tint="@color/icon_warning" />
```
antonijzelinskij, line 83: Is this icon valid? Looking like temporary stub.
Do we have any design for this feature?

antonijzelinskij (changes requested):

common/src/main/res/values-es/strings.xml
Leemo94: Done — removed the 7 `dapp_search_staking_*` and `dapp_staking_competitor_*` entries from all 13 non-English locale files (values-es, values-fr-rFR, values-hu, values-in, values-it, values-ja, values-ko, values-pl, values-pt, values-ru, values-tr, values-vi, values-zh-rCN). English source in `values/strings.xml` kept unchanged for the localization service.

feature-dapp-impl/src/main/java/io/novafoundation/nova/feature_dapp_impl/domain/search/StakingKeywordMatcher.kt
Leemo94: Done — converted all 123 \\uXXXX escape sequences in MULTILINGUAL_KEYWORDS and CJK_KEYWORDS to literal unicode characters. Zero \\u-escapes remain in the file.

feature-dapp-impl/src/main/java/io/novafoundation/nova/feature_dapp_impl/domain/search/StakingKeywordMatcher.kt
Leemo94: Same as above — all \\uXXXX escapes in the file have been replaced with literal unicode characters.

app/src/main/java/io/novafoundation/nova/app/root/navigation/navigators/dApp/DAppNavigator.kt
Leemo94: Done in `47c147956`. `navigateToStaking()` now uses `navigationBuilder().action(R.id.action_open_split_screen).navigateInRoot()` + `stakingDashboardRouter.openStakingDashboard()`, matching the canonical pattern in `RelayStakingNavigator.openStakingDashboard()` (lines 285-292). The address-bar-search case is now safe — `action_open_split_screen` pops the entire root nav stack (its `popUpTo="@id/root_nav_graph"` with `popUpToInclusive="true"` clears any browser sub-screen including the in-browser search), so `openStakingDashboard` always runs from the main-screen context.

app/src/main/java/io/novafoundation/nova/app/root/navigation/navigators/staking/StakingDashboardNavigator.kt
Leemo94: Done in `47c147956`. File fully reverted to upstream — confirmed `performNavigationOrDelay(R.id.action_open_staking)` works correctly with the existing `setupWithNavController` wiring (menu item and destination both share the `staking_dashboard_graph` ID, so the bottom-nav highlight syncs automatically). Dropped the added `bottomNavigationView` field, `setBottomNavigationView(view:)`, `pendingSelectedItemId`, and the modified `openStakingDashboard` body.

app/src/main/java/io/novafoundation/nova/app/root/presentation/main/MainFragment.kt
Leemo94: Done in `47c147956`. The `stakingDashboardNavigator.setBottomNavigationView(...)` line is removed. `setupWithNavController(navController!!)` on the line above is sufficient.

feature-dapp-impl/src/main/java/io/novafoundation/nova/feature_dapp_impl/domain/browser/StakingCompetitorDomains.kt
Leemo94: Done in `9d2e8e034`. Migrated to `StakingCompetitorDomainsRepository` (Retrofit interface + impl with in-memory `@Volatile List<String>` cache + fail-open semantics) fed by `BuildConfig.STAKING_COMPETITORS_URL` (dev → `staking_competitors_dev.json`, release → `staking_competitors.json`). DI via a new `StakingCompetitorDomainsModule`. Sync triggered at `DAppBrowserViewModel.init` and `DAppSearchViewModel.init`.

Companion nova-utils PR: novasamatech/nova-utils#4343.

Cache is sync-readable rather than `suspend` because `shouldOverrideUrlLoading()` requires a synchronous Boolean return. On a failed fetch (or before the first sync completes), `isStakingCompetitor` returns false → no notice fires → navigation proceeds normally. The existing `watchStakingCompetitorInitialLoad` Flow catches the URL post-load if the cache populates while the user is still on the competitor page.

feature-dapp-impl/src/main/java/io/novafoundation/nova/feature_dapp_impl/presentation/browser/main/DAppBrowserFragment.kt
Leemo94: Done in `fac39766c`. The Fragment no longer mutates `binder.dappBrowserStakingWarning*` visibility directly. Replaced `LiveData<Boolean>` with `LiveData<StakingWarningState>` (a data class with `visible` + `advancedRevealed` flags). The Fragment's observer reads the state object and binds visibility declaratively; new `onStakingWarningAdvancedClicked()` ViewModel method moves the advanced/continue toggle off the Fragment.

feature-dapp-impl/src/main/java/io/novafoundation/nova/feature_dapp_impl/presentation/browser/main/DAppBrowserFragment.kt
Leemo94: Done in `9d2e8e034` (provider) + `fac39766c` (VM-side check). The fragment's `shouldOverrideUrlLoading` is now `if (viewModel.onStakingCompetitorIntercepted(url)) { return true }` — exactly the shape you proposed. The `isStakingCompetitor` check moved inside `onStakingCompetitorIntercepted` on the ViewModel, which delegates to the injected `StakingCompetitorDomainsRepository`. The static `StakingCompetitorDomains` object is deleted entirely; the Fragment loses both the call and the import.

feature-dapp-impl/src/main/res/layout/fragment_dapp_browser.xml
Leemo94: The current icon (`@drawable/ic_info_24`) reads as advisory rather than alarming, which matches the soft-nudge intent of the notice (the user always has Continue and Go-to-Staking — it's not a block). Happy with it as-is from our side; if the dev team prefers a different icon, please propose a specific alternative.

PR #2274, opened 2026-04-08
Add Nova service fee for Hydration swaps

Leemo94:
⚠️ Requires ToS update

Introduce a 0.85% commission on Hydration swap output, calculated from the post-slippage minimum amount and transferred to a Nova fee account via a batched extrinsic (batchAll with swap + transfer).

- Generic BundleExtraActions hook on AtomicSwapOperation keeps the AssetExchange layer fee-agnostic; fee logic lives in SwapService
- Rate display bakes in the fee (MetaMask-style UX)
- Conditional "Includes 0.85% Nova fee." disclaimer on swap screens
- Rate tooltip is fee-aware for Hydration swaps, generic otherwise
- "Network fees" label replaces "Total fee"
- Price impact intentionally excludes the commission
- Cross-chain swaps: commission applies to the Hydration segment only
- NovaSwapCommission.kt constants, SwapQuoteNovaFee.kt extensions
- Dynamic ORML module resolution via runtime.metadata.firstExistingCall()
- Fee transferred via Tokens/Currencies.transfer for ORML tokens, Balances.transferKeepAlive for native HDX
- Transaction history filter for fee transfers tracked as TODO
- 13 language translations for fee-related UI strings

### DIFF
--- a/common/src/main/java/io/novafoundation/nova/common/view/bottomSheet/description/DescriptionBottomSheet.kt
+++ b/common/src/main/java/io/novafoundation/nova/common/view/bottomSheet/description/DescriptionBottomSheet.kt
@@ -11,20 +11,17 @@ import io.novafoundation.nova.common.view.bottomSheet.BaseBottomSheet
 class DescriptionBottomSheet(
     context: Context,
     val titleRes: Int,
-    val descriptionRes: Int,
-    val descriptionText: String? = null
+    val descriptionText: String
 ) : BaseBottomSheet<BottomSheetDescriptionBinding>(context, R.style.BottomSheetDialog), WithContextExtensions by WithContextExtensions(context) {
+
+    constructor(context: Context, titleRes: Int, descriptionRes: Int) :
+        this(context, titleRes, context.getString(descriptionRes))
 
     override val binder: BottomSheetDescriptionBinding = BottomSheetDescriptionBinding.inflate(LayoutInflater.from(context))
 
     override fun onCreate(savedInstanceState: Bundle?) {
         super.onCreate(savedInstanceState)
         binder.sheetDescriptionTitle.setText(titleRes)
-
-        if (descriptionText != null) {
-            binder.sheetDescriptionDetails.text = descriptionText
-        } else {
-            binder.sheetDescriptionDetails.setText(descriptionRes)
-        }
+        binder.sheetDescriptionDetails.text = descriptionText
     }
 }
--- a/common/src/main/java/io/novafoundation/nova/common/view/bottomSheet/description/DescriptionBottomSheet.kt
+++ b/common/src/main/java/io/novafoundation/nova/common/view/bottomSheet/description/DescriptionBottomSheet.kt
@@ -11,7 +11,8 @@ import io.novafoundation.nova.common.view.bottomSheet.BaseBottomSheet
 class DescriptionBottomSheet(
     context: Context,
     val titleRes: Int,
-    val descriptionRes: Int
+    val descriptionRes: Int,
+    val descriptionText: String? = null
 ) : BaseBottomSheet<BottomSheetDescriptionBinding>(context, R.style.BottomSheetDialog), WithContextExtensions by WithContextExtensions(context) {
 
     override val binder: BottomSheetDescriptionBinding = BottomSheetDescriptionBinding.inflate(LayoutInflater.from(context))
@@ -19,6 +20,11 @@ class DescriptionBottomSheet(
     override fun onCreate(savedInstanceState: Bundle?) {
         super.onCreate(savedInstanceState)
         binder.sheetDescriptionTitle.setText(titleRes)
-        binder.sheetDescriptionDetails.setText(descriptionRes)
+
+        if (descriptionText != null) {
+            binder.sheetDescriptionDetails.text = descriptionText
+        } else {
+            binder.sheetDescriptionDetails.setText(descriptionRes)
+        }
     }
 }
--- a/common/src/main/res/values-fr-rFR/strings.xml
+++ b/common/src/main/res/values-fr-rFR/strings.xml
@@ -1805,6 +1805,9 @@
     <string name="swap_price_difference_title">Différence de prix</string>
     <string name="swap_price_impact" formatted="false">%s ≈ %s</string>
     <string name="swap_rate_description">Taux de change entre deux crypto-monnaies différentes. Il représente combien de crypto-monnaie vous pouvez obtenir en échange d\'un certain montant d\'une autre crypto-monnaie.</string>
+    <string name="swap_rate_includes_fee_description">Le meilleur taux trouvé, incluant des frais Nova Wallet de %s%%.</string>
+    <string name="swap_network_fees_label">Frais de réseau</string>
+    <string name="swap_nova_fee_disclaimer">Inclut des frais Nova de %s%%.</string>
     <string name="swap_rate_title">Taux</string>
     <string name="swap_rate_was_updated_failure_message">Ancien taux : %1$st∻%2$s.\nNouveau taux : %1$st∻%3$s</string>
     <string name="swap_rate_was_updated_failure_title">Le taux d\'échange a été mis à jour</string>
--- a/common/src/main/res/values-fr-rFR/strings.xml
+++ b/common/src/main/res/values-fr-rFR/strings.xml
@@ -1805,7 +1805,6 @@
     <string name="swap_price_difference_title">Différence de prix</string>
     <string name="swap_price_impact" formatted="false">%s ≈ %s</string>
     <string name="swap_rate_description">Taux de change entre deux crypto-monnaies différentes. Il représente combien de crypto-monnaie vous pouvez obtenir en échange d\'un certain montant d\'une autre crypto-monnaie.</string>
-    <string name="swap_rate_includes_fee_description">Le meilleur taux trouvé, incluant des frais Nova Wallet de %s%%.</string>
     <string name="swap_network_fees_label">Frais de réseau</string>
     <string name="swap_nova_fee_disclaimer">Inclut des frais Nova de %s%%.</string>
     <string name="swap_rate_title">Taux</string>
--- a/feature-assets/src/main/java/io/novafoundation/nova/feature_assets/data/repository/TransactionHistoryRepository.kt
+++ b/feature-assets/src/main/java/io/novafoundation/nova/feature_assets/data/repository/TransactionHistoryRepository.kt
@@ -1,6 +1,5 @@
 package io.novafoundation.nova.feature_assets.data.repository
 
-import io.novafoundation.nova.common.address.AccountIdKey
 import io.novafoundation.nova.common.data.model.DataPage
 import io.novafoundation.nova.common.data.model.PageOffset
 import io.novafoundation.nova.common.utils.Filter
@@ -8,8 +7,6 @@ import io.novafoundation.nova.common.utils.applyFilters
 import io.novafoundation.nova.core_db.dao.OperationDao
 import io.novafoundation.nova.core_db.model.operation.OperationBaseLocal
 import io.novafoundation.nova.core_db.model.operation.OperationJoin
-import io.novafoundation.nova.feature_account_api.domain.account.system.AccountSystemAccountMatcher
-import io.novafoundation.nova.feature_account_api.domain.account.system.CompoundSystemAccountMatcher
 import io.novafoundation.nova.feature_account_api.domain.account.system.SystemAccountMatcher
 import io.novafoundation.nova.feature_assets.data.mappers.mapOperationLocalToOperation
 import io.novafoundation.nova.feature_assets.data.mappers.mapOperationToOperationLocalDb
@@ -17,8 +14,7 @@ import io.novafoundation.nova.feature_currency_api.domain.model.Currency
 import io.novafoundation.nova.feature_staking_api.data.mythos.MythosMainPotMatcherFactory
 import io.novafoundation.nova.feature_staking_api.data.nominationPools.pool.PoolAccountDerivation
 import io.novafoundation.nova.feature_staking_api.data.nominationPools.pool.poolRewardAccountMatcher
-import io.novafoundation.nova.feature_swap_api.domain.model.HydrationSystemAccounts
-import io.novafoundation.nova.feature_swap_api.domain.model.NovaSwapCommission
+import io.novafoundation.nova.feature_swap_api.data.history.HydrationSwapTransferFilterFactory
 import io.novafoundation.nova.feature_wallet_api.data.network.blockhain.assets.AssetSourceRegistry
 import io.novafoundation.nova.feature_wallet_api.data.network.blockhain.assets.history.AssetHistory
 import io.novafoundation.nova.feature_wallet_api.data.repository.CoinPriceRepository
@@ -29,7 +25,6 @@ import io.novafoundation.nova.feature_wallet_api.domain.model.Operation
 import io.novafoundation.nova.feature_wallet_api.domain.model.findNearestCoinRate
 import io.novafoundation.nova.runtime.ext.accountIdOrNull
 import io.novafoundation.nova.runtime.ext.addressOf
-import io.novafoundation.nova.runtime.ext.hydraDxSupported
 import io.novafoundation.nova.runtime.multiNetwork.chain.model.Chain
 import io.novasama.substrate_sdk_android.runtime.AccountId
 import kotlinx.coroutines.Dispatchers
@@ -73,7 +68,8 @@ class RealTransactionHistoryRepository(
     private val operationDao: OperationDao,
     private val poolAccountDerivation: PoolAccountDerivation,
     private val mythosMainPotMatcherFactory: MythosMainPotMatcherFactory,
-    private val coinPriceRepository: CoinPriceRepository
+    private val hydrationSwapTransferFilterFactory: HydrationSwapTransferFilterFactory,
+    private val coinPriceRepository: CoinPriceRepository,
 ) : TransactionHistoryRepository {
 
     override suspend fun syncOperationsFirstPage(
@@ -189,21 +185,11 @@ class RealTransactionHistoryRepository(
             matcher?.let { IgnoreTransfersFromSystemAccount(it, chain) }
         }
 
-        val hydrationSwapSystemAccountFilter = if (chain.swap.hydraDxSupported()) {
-            val matcher = CompoundSystemAccountMatcher(
-                AccountSystemAccountMatcher(AccountIdKey(NovaSwapCommission.feeAccountId)),
-                AccountSystemAccountMatcher(AccountIdKey(HydrationSystemAccounts.routerAccountId))
-            )
-            IgnoreTransfersInvolvingSystemAccount(matcher, chain)
-        } else {
-            null
-        }
-
         return listOfNotNull(
             IgnoreUnsafeOperations(this),
             systemAccountFilterCreator(poolAccountDerivation.poolRewardAccountMatcher(chain.id)),
             systemAccountFilterCreator(mythosMainPotMatcherFactory.create(chainAsset)),
-            hydrationSwapSystemAccountFilter
+            hydrationSwapTransferFilterFactory.create(chain),
         )
     }
 
@@ -224,24 +210,6 @@ class RealTransactionHistoryRepository(
 
         override fun shouldInclude(model: Operation): Boolean {
             return assetsHistory.isOperationSafe(model)
-        }
-    }
-
-    private class IgnoreTransfersInvolvingSystemAccount(
-        private val systemAccountMatcher: SystemAccountMatcher,
-        private val chain: Chain
-    ) : Filter<Operation> {
-
-        override fun shouldInclude(model: Operation): Boolean {
-            val operationType = model.type as? Operation.Type.Transfer ?: return true
-
-            return !chain.matchesSystemAccount(operationType.sender) &&
-                !chain.matchesSystemAccount(operationType.receiver)
-        }
-
-        private fun Chain.matchesSystemAccount(address: String): Boolean {
-            val accountId = accountIdOrNull(address) ?: return false
-            return systemAccountMatcher.isSystemAccount(accountId)
         }
     }
 }
--- a/feature-assets/src/main/java/io/novafoundation/nova/feature_assets/data/repository/TransactionHistoryRepository.kt
+++ b/feature-assets/src/main/java/io/novafoundation/nova/feature_assets/data/repository/TransactionHistoryRepository.kt
@@ -1,5 +1,6 @@
 package io.novafoundation.nova.feature_assets.data.repository
 
+import io.novafoundation.nova.common.address.AccountIdKey
 import io.novafoundation.nova.common.data.model.DataPage
 import io.novafoundation.nova.common.data.model.PageOffset
 import io.novafoundation.nova.common.utils.Filter
@@ -7,6 +8,8 @@ import io.novafoundation.nova.common.utils.applyFilters
 import io.novafoundation.nova.core_db.dao.OperationDao
 import io.novafoundation.nova.core_db.model.operation.OperationBaseLocal
 import io.novafoundation.nova.core_db.model.operation.OperationJoin
+import io.novafoundation.nova.feature_account_api.domain.account.system.AccountSystemAccountMatcher
+import io.novafoundation.nova.feature_account_api.domain.account.system.CompoundSystemAccountMatcher
 import io.novafoundation.nova.feature_account_api.domain.account.system.SystemAccountMatcher
 import io.novafoundation.nova.feature_assets.data.mappers.mapOperationLocalToOperation
 import io.novafoundation.nova.feature_assets.data.mappers.mapOperationToOperationLocalDb
@@ -14,6 +17,8 @@ import io.novafoundation.nova.feature_currency_api.domain.model.Currency
 import io.novafoundation.nova.feature_staking_api.data.mythos.MythosMainPotMatcherFactory
 import io.novafoundation.nova.feature_staking_api.data.nominationPools.pool.PoolAccountDerivation
 import io.novafoundation.nova.feature_staking_api.data.nominationPools.pool.poolRewardAccountMatcher
+import io.novafoundation.nova.feature_swap_api.domain.model.HydrationSystemAccounts
+import io.novafoundation.nova.feature_swap_api.domain.model.NovaSwapCommission
 import io.novafoundation.nova.feature_wallet_api.data.network.blockhain.assets.AssetSourceRegistry
 import io.novafoundation.nova.feature_wallet_api.data.network.blockhain.assets.history.AssetHistory
 import io.novafoundation.nova.feature_wallet_api.data.repository.CoinPriceRepository
@@ -24,6 +29,7 @@ import io.novafoundation.nova.feature_wallet_api.domain.model.Operation
 import io.novafoundation.nova.feature_wallet_api.domain.model.findNearestCoinRate
 import io.novafoundation.nova.runtime.ext.accountIdOrNull
 import io.novafoundation.nova.runtime.ext.addressOf
+import io.novafoundation.nova.runtime.ext.hydraDxSupported
 import io.novafoundation.nova.runtime.multiNetwork.chain.model.Chain
 import io.novasama.substrate_sdk_android.runtime.AccountId
 import kotlinx.coroutines.Dispatchers
@@ -183,10 +189,21 @@ class RealTransactionHistoryRepository(
             matcher?.let { IgnoreTransfersFromSystemAccount(it, chain) }
         }
 
+        val hydrationSwapSystemAccountFilter = if (chain.swap.hydraDxSupported()) {
+            val matcher = CompoundSystemAccountMatcher(
+                AccountSystemAccountMatcher(AccountIdKey(NovaSwapCommission.feeAccountId)),
+                AccountSystemAccountMatcher(AccountIdKey(HydrationSystemAccounts.routerAccountId))
+            )
+            IgnoreTransfersInvolvingSystemAccount(matcher, chain)
+        } else {
+            null
+        }
+
         return listOfNotNull(
             IgnoreUnsafeOperations(this),
             systemAccountFilterCreator(poolAccountDerivation.poolRewardAccountMatcher(chain.id)),
-            systemAccountFilterCreator(mythosMainPotMatcherFactory.create(chainAsset))
+            systemAccountFilterCreator(mythosMainPotMatcherFactory.create(chainAsset)),
+            hydrationSwapSystemAccountFilter
         )
     }
 
@@ -207,6 +224,24 @@ class RealTransactionHistoryRepository(
 
         override fun shouldInclude(model: Operation): Boolean {
             return assetsHistory.isOperationSafe(model)
+        }
+    }
+
+    private class IgnoreTransfersInvolvingSystemAccount(
+        private val systemAccountMatcher: SystemAccountMatcher,
+        private val chain: Chain
+    ) : Filter<Operation> {
+
+        override fun shouldInclude(model: Operation): Boolean {
+            val operationType = model.type as? Operation.Type.Transfer ?: return true
+
+            return !chain.matchesSystemAccount(operationType.sender) &&
+                !chain.matchesSystemAccount(operationType.receiver)
+        }
+
+        private fun Chain.matchesSystemAccount(address: String): Boolean {
+            val accountId = accountIdOrNull(address) ?: return false
+            return systemAccountMatcher.isSystemAccount(accountId)
         }
     }
 }
--- a/feature-swap-api/src/main/java/io/novafoundation/nova/feature_swap_api/presentation/view/bottomSheet/description/DescriptionBottomSheetLauncherExt.kt
+++ b/feature-swap-api/src/main/java/io/novafoundation/nova/feature_swap_api/presentation/view/bottomSheet/description/DescriptionBottomSheetLauncherExt.kt
@@ -6,10 +6,9 @@ import io.novafoundation.nova.feature_swap_api.R
 
 fun DescriptionBottomSheetLauncher.launchSwapRateDescription(
     resourceManager: ResourceManager,
-    includesNovaFee: Boolean,
-    feePercentDisplay: String
+    feePercentDisplay: String?
 ) {
-    if (includesNovaFee) {
+    if (feePercentDisplay != null) {
         val description = resourceManager.getString(R.string.swap_rate_includes_fee_description, feePercentDisplay)
         launchDescriptionBottomSheet(
             titleRes = R.string.swap_rate_title,
--- a/feature-swap-api/src/main/java/io/novafoundation/nova/feature_swap_api/presentation/view/bottomSheet/description/DescriptionBottomSheetLauncherExt.kt
+++ b/feature-swap-api/src/main/java/io/novafoundation/nova/feature_swap_api/presentation/view/bottomSheet/description/DescriptionBottomSheetLauncherExt.kt
@@ -1,13 +1,26 @@
 package io.novafoundation.nova.feature_swap_api.presentation.view.bottomSheet.description
 
+import io.novafoundation.nova.common.resources.ResourceManager
 import io.novafoundation.nova.common.view.bottomSheet.description.DescriptionBottomSheetLauncher
 import io.novafoundation.nova.feature_swap_api.R
 
-fun DescriptionBottomSheetLauncher.launchSwapRateDescription() {
-    launchDescriptionBottomSheet(
-        titleRes = R.string.swap_rate_title,
-        descriptionRes = R.string.swap_rate_description
-    )
+fun DescriptionBottomSheetLauncher.launchSwapRateDescription(
+    resourceManager: ResourceManager,
+    includesNovaFee: Boolean,
+    feePercentDisplay: String
+) {
+    if (includesNovaFee) {
+        val description = resourceManager.getString(R.string.swap_rate_includes_fee_description, feePercentDisplay)
+        launchDescriptionBottomSheet(
+            titleRes = R.string.swap_rate_title,
+            descriptionText = description
+        )
+    } else {
+        launchDescriptionBottomSheet(
+            titleRes = R.string.swap_rate_title,
+            descriptionRes = R.string.swap_rate_description
+        )
+    }
 }
 
 fun DescriptionBottomSheetLauncher.launchPriceDifferenceDescription() {
--- a/feature-swap-api/src/main/java/io/novafoundation/nova/feature_swap_api/presentation/view/bottomSheet/description/DescriptionBottomSheetLauncherExt.kt
+++ b/feature-swap-api/src/main/java/io/novafoundation/nova/feature_swap_api/presentation/view/bottomSheet/description/DescriptionBottomSheetLauncherExt.kt
@@ -4,20 +4,26 @@ import io.novafoundation.nova.common.resources.ResourceManager
 import io.novafoundation.nova.common.view.bottomSheet.description.DescriptionBottomSheetLauncher
 import io.novafoundation.nova.feature_swap_api.R
 
+sealed class SwapRateDescriptionMode {
+
+    object Default : SwapRateDescriptionMode()
+
+    data class IncludesFee(val feePercentDisplay: String) : SwapRateDescriptionMode()
+}
+
 fun DescriptionBottomSheetLauncher.launchSwapRateDescription(
     resourceManager: ResourceManager,
-    feePercentDisplay: String?
+    mode: SwapRateDescriptionMode,
 ) {
-    if (feePercentDisplay != null) {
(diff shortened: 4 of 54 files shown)


common/src/main/java/io/novafoundation/nova/common/view/bottomSheet/description/DescriptionBottomSheet.kt
```diff
@@ -11,14 +11,20 @@ import io.novafoundation.nova.common.view.bottomSheet.BaseBottomSheet
 class DescriptionBottomSheet(
     context: Context,
     val titleRes: Int,
-    val descriptionRes: Int
+    val descriptionRes: Int,
+    val descriptionText: String? = null
```
valentunn: The easier way is to always accept `descriptionText` in the primary constructor and define a secondary constructor that accepts `descriptionRes` and passes `context.getString(descriptionRes)` to the primary

feature-swap-api/src/main/java/io/novafoundation/nova/feature_swap_api/presentation/view/bottomSheet/description/DescriptionBottomSheetLauncherExt.kt
```diff
@@ -1,13 +1,26 @@
...
 
-fun DescriptionBottomSheetLauncher.launchSwapRateDescription() {
-    launchDescriptionBottomSheet(
-        titleRes = R.string.swap_rate_title,
-        descriptionRes = R.string.swap_rate_description
-    )
+fun DescriptionBottomSheetLauncher.launchSwapRateDescription(
+    resourceManager: ResourceManager,
+    includesNovaFee: Boolean,
+    feePercentDisplay: String
```
valentunn: Just pass `feePercentDisplay: String?` no need for a separate `includesNovaFee` flag

feature-swap-impl/src/main/java/io/novafoundation/nova/feature_swap_impl/domain/swap/RealSwapService.kt
```diff
@@ -282,6 +317,88 @@ internal class RealSwapService(
...
+    /**
+     * Creates a [BundleExtraActions] lambda that adds a transfer call for the commission.
+     * Uses Balances.transfer_keep_alive for native assets.
+     * For ORML tokens, dynamically resolves between Currencies and Tokens modules
+     * using [firstExistingCall], matching the approach in [OrmlAssetTransfers].
+     */
+    private fun buildCommissionTransferCall(
+        assetOut: Chain.Asset,
+        commissionAmount: Balance
+    ): BundleExtraActions {
```
valentunn: Use `AssetTransfers` via `AssetSourceRegistry.sourceFor` to construct a transferCall
Note that it does not support constructing `GenericCall` yet - we should create a new method in `AssetTransfers`. In implementations re-use the existing logic for constructing calls that are within those types

feature-swap-impl/src/main/java/io/novafoundation/nova/feature_swap_impl/domain/swap/RealSwapService.kt
```diff
@@ -264,6 +281,22 @@ internal class RealSwapService(
         }
     }
 
+    /**
+     * Finds the index of the last segment whose output chain supports HydraDx swaps.
+     * Returns -1 if no segment runs on a Hydra chain.
+     */
+    private suspend fun findLastHydraSegmentIndex(segments: List<SwapFee.SwapSegment>): Int {
+        var lastIndex = -1
+        segments.forEachIndexed { index, (_, operation) ->
```
valentunn: `segments.indexOf { <search logic> }  `

feature-swap-impl/src/main/java/io/novafoundation/nova/feature_swap_impl/domain/swap/RealSwapService.kt
```diff
@@ -264,6 +281,22 @@ internal class RealSwapService(
...
+            val chainId = operation.assetOut.chainId
+            val chain = chainRegistry.getChain(chainId)
+            if (chain.swap.hydraDxSupported()) {
```
valentunn: This will not work correctly if the last segment on hydration is cross-chain transfer

We can use more direct detection via `segment.operation is HydraDxOperation`. You can make `HydraDxAssetExchange` non private if needed

feature-swap-impl/src/main/java/io/novafoundation/nova/feature_swap_impl/domain/swap/RealSwapService.kt
```diff
@@ -282,6 +317,88 @@ internal class RealSwapService(
...
+        operation: AtomicSwapOperation,
+        actualSwapLimit: SwapLimit
+    ): Balance {
+        val chainId = operation.assetOut.chainId
+        val chain = chainRegistry.getChain(chainId)
+
+        if (!chain.swap.hydraDxSupported()) return BigInteger.ZERO
+
+        val commissionAmount = NovaSwapCommission.feeAmount(actualSwapLimit.amountOutMin)
+        return if (commissionAmount > BigInteger.ZERO) commissionAmount else BigInteger.ZERO
```
valentunn: commissionAmount.atLeastZero()

feature-swap-impl/src/main/java/io/novafoundation/nova/feature_swap_impl/domain/swap/SwapQuoteNovaFee.kt
```diff
@@ -0,0 +1,41 @@
...
+import java.math.BigDecimal
+
+/**
+ * Checks whether any segment in the quoted path involves a swap on a Hydration chain.
+ */
+fun SwapQuote.involvesHydraSwap(chainsById: ChainsById): Boolean {
+    return quotedPath.path.any { quotedEdge ->
+        val chainId = quotedEdge.edge.from.chainId
+        val chain = chainsById[chainId] ?: return@any false
+        chain.swap.hydraDxSupported()
```
valentunn: Here as well: use direct `edge is HydraDxQuotableEdge` check

feature-swap-impl/src/main/java/io/novafoundation/nova/feature_swap_impl/domain/swap/SwapQuoteNovaFee.kt
```diff
@@ -0,0 +1,41 @@
...
+    }
+}
+
+/**
+ * Returns the display amount out - after subtracting the Nova commission fee for Hydra swaps.
+ * For non-Hydra swaps, returns the original amount out.
+ */
+fun SwapQuote.displayAmountOut(chainsById: ChainsById): Balance {
```
valentunn: We should rather override amountOut when constructing `SwapQuote`. So the logic always works with correct (post-fees) amount

feature-wallet-impl/src/main/java/io/novafoundation/nova/feature_wallet_impl/data/network/blockchain/assets/history/SubstrateAssetHistory.kt
```diff
@@ -147,6 +147,11 @@ abstract class SubstrateAssetHistory(
 
         val assetsBySubQueryId = chain.assetsBySubQueryId()
 
+        // TODO: Filter out commission transfers to the Nova fee account (NovaSwapCommission.FEE_ACCOUNT_HEX)
+        //  so they don't appear as separate transfer entries in the user's transaction history.
+        //  The fee account address varies by chain encoding, so the filter should compare account IDs
+        //  rather than formatted addresses. This requires making the fee account ID accessible from
+        //  a shared module (e.g. feature-wallet-api) to avoid circular dependencies with feature-swap-impl.
```
valentunn: We can easily achieve that by creating a new filter in `createTransactionFilters`

common/src/main/java/io/novafoundation/nova/common/view/bottomSheet/description/DescriptionBottomSheet.kt
Leemo94: Done in b1354236d. Primary ctor takes `descriptionText: String`; secondary takes `descriptionRes: Int` and resolves it via `context.getString`.                                                                                                             `observeDescription` branches on `event.descriptionText` nullness so each call site picks one path.

feature-swap-api/src/main/java/io/novafoundation/nova/feature_swap_api/presentation/view/bottomSheet/description/DescriptionBottomSheetLauncherExt.kt
Leemo94: Done in b1354236d. Signature is now `launchSwapRateDescription(resourceManager,                                                                                                    feePercentDisplay: String?)`. Null hides the fee-aware variant; non-null                                                                                                         drives the includes-fee message. Callers pass `if (includesFee) NovaSwapCommission.FEE_PERCENT_DISPLAY else null`.

feature-swap-impl/src/main/java/io/novafoundation/nova/feature_swap_impl/domain/swap/RealSwapService.kt
Leemo94: Done in 93f69bb6f. Added `AssetTransfers.constructTransferCall(transfer: AssetTransferBase): GenericCall.Instance` to the interface; default impl throws so subclasses opt in. `NativeAssetTransfers` and `OrmlAssetTransfers` implement it via `composeCall` and their existing `transferFunctions` resolution. `buildCommissionActions` now builds an `AssetTransferBase` and goes through `assetSourceRegistry.sourceFor(assetOut).transfers.constructTransferCall(...)`. The hand-rolled `Balances.transfer_keep_alive` / `Currencies.transfer` branching is gone.

feature-swap-impl/src/main/java/io/novafoundation/nova/feature_swap_impl/domain/swap/RealSwapService.kt
Leemo94: Done in 655c0b755. `findLastHydraSegmentIndex` is now `segments.indexOfLast { (_, operation) -> operation is HydraDxOperation }`. The type-check switch from the next thread is the same rewrite.

feature-swap-impl/src/main/java/io/novafoundation/nova/feature_swap_impl/domain/swap/RealSwapService.kt
Leemo94: Done in 655c0b755. Made `HydraDxOperation` `internal` (was `private`) and switched the gate to the type check. The previous chain-id approach was overmatching cross-chain transfer edges to/from Hydration as if they were swaps, which would have skipped the commission on a multi-hop bundle whose last leg is a transfer off Hydra.

feature-swap-impl/src/main/java/io/novafoundation/nova/feature_swap_impl/domain/swap/RealSwapService.kt
Leemo94: Done in b1354236d. `getCommissionAmount` is now `NovaSwapCommission.feeAmount(actualSwapLimit.amountOutMin).atLeastZero()`. While I was in here I also clamped the post-segment correction `(executionCorrection.actualReceivedAmount - commissionAmount).atLeastZero()` defensively so a dry-run/actual mismatch can't produce a negative balance view downstream.

feature-swap-impl/src/main/java/io/novafoundation/nova/feature_swap_impl/domain/swap/SwapQuoteNovaFee.kt
Leemo94: Done in 655c0b755. `involvesHydraSwap()` is now `quotedPath.path.any { it.edge is HydraDxQuotableEdge }`.

feature-swap-impl/src/main/java/io/novafoundation/nova/feature_swap_impl/domain/swap/SwapQuoteNovaFee.kt
Leemo94: Done in e534f8d85. `SwapQuote.amountOut` is post-commission at construction in `RealSwapService.createQuote` for Hydra-involving paths, so every consumer (UI, validations, rate display) sees the value the user actually receives. `displayAmountOut` and `displaySwapRate` are gone; call sites use the existing `swapRate()` extension on `SwapQuote`. The pool's gross quote is still reachable via `quotedPath`'s last-segment quote for the few places that need it (notably segment-level `SwapLimit` construction, which drives the on-chain extrinsic args and must remain gross).

One deliberate carve-out worth flagging: `priceImpact` is still computed against the gross amount (`amountOutGross`). Computing it against the net amount inflates the reading by the commission percentage and reproduces a bug we hit in an earlier iteration where the displayed slippage looked ~0.85pp worse than the actual pool execution. Happy to revisit if you'd prefer the post-fee version.

feature-wallet-impl/src/main/java/io/novafoundation/nova/feature_wallet_impl/data/network/blockchain/assets/history/SubstrateAssetHistory.kt
Leemo94: Done in 93f69bb6f, but landed in `RealTransactionHistoryRepository.createTransactionFilters` (feature-assets) rather than `SubstrateAssetHistory`, because that's where the existing `systemAccountFilterCreator` lives alongside the pool-reward and Mythos main-pot filters. The new `IgnoreTransfersInvolvingSystemAccount` drops any transfer whose sender or receiver matches a Hydration system account, gated by `chain.swap.hydraDxSupported()`.

Heads-up on scope: the matcher is a `CompoundSystemAccountMatcher` covering two accounts, not just the commission recipient. The Hydration router pallet's substrate-derived account (`modlrouterex`, in `HydrationSystemAccounts.kt`) shows up because the router temporarily holds the user's funds during swap execution, so the indexer surfaces a user→router and router→user pair on top of the actual swap entry. Hiding both keeps history showing only the swap. Happy to split the router piece into a separate change if you'd rather keep this PR narrowly scoped to the fee-account filter.

Leemo94:
Screenshots showing UI changes:

swap 1
swap 2
swap 3

common/src/main/res/values-fr-rFR/strings.xml
```diff
@@ -1805,6 +1805,9 @@
     <string name="swap_price_difference_title">Différence de prix</string>
     <string name="swap_price_impact" formatted="false">%s ≈ %s</string>
     <string name="swap_rate_description">Taux de change entre deux crypto-monnaies différentes. Il représente combien de crypto-monnaie vous pouvez obtenir en échange d\'un certain montant d\'une autre crypto-monnaie.</string>
+    <string name="swap_rate_includes_fee_description">Le meilleur taux trouvé, incluant des frais Nova Wallet de %s%%.</string>
```
antonijzelinskij: blocking

We don't translate string resources like this. We always use Localisation services for that right before release.

So please remove everything besides english strings

feature-assets/src/main/java/io/novafoundation/nova/feature_assets/data/repository/TransactionHistoryRepository.kt
```diff
@@ -183,10 +189,21 @@ class RealTransactionHistoryRepository(
...
 
+        val hydrationSwapSystemAccountFilter = if (chain.swap.hydraDxSupported()) {
+            val matcher = CompoundSystemAccountMatcher(
+                AccountSystemAccountMatcher(AccountIdKey(NovaSwapCommission.feeAccountId)),
+                AccountSystemAccountMatcher(AccountIdKey(HydrationSystemAccounts.routerAccountId))
+            )
+            IgnoreTransfersInvolvingSystemAccount(matcher, chain)
+        } else {
+            null
+        }
```
antonijzelinskij: Let's use factory for that case like mythosMainPotMatcherFactory for example

feature-assets/src/main/java/io/novafoundation/nova/feature_assets/data/repository/TransactionHistoryRepository.kt
```diff
@@ -209,4 +226,22 @@ class RealTransactionHistoryRepository(
             return assetsHistory.isOperationSafe(model)
         }
     }
+
+    private class IgnoreTransfersInvolvingSystemAccount(
```
antonijzelinskij: Let's put it separately. Right in created factory would be great

feature-swap-api/src/main/java/io/novafoundation/nova/feature_swap_api/presentation/view/bottomSheet/description/DescriptionBottomSheetLauncherExt.kt
```diff
@@ -1,13 +1,25 @@
...
 import io.novafoundation.nova.feature_swap_api.R
 
-fun DescriptionBottomSheetLauncher.launchSwapRateDescription() {
-    launchDescriptionBottomSheet(
-        titleRes = R.string.swap_rate_title,
-        descriptionRes = R.string.swap_rate_description
-    )
+fun DescriptionBottomSheetLauncher.launchSwapRateDescription(
+    resourceManager: ResourceManager,
+    feePercentDisplay: String?
```
antonijzelinskij: Not blocking but better to have for better code readness 
Would be good to use sealed class like class FeeDisplay.FeeDescription(feeAmount: String) and object FeeDisplay.Default

feature-swap-impl/src/main/java/io/novafoundation/nova/feature_swap_impl/domain/swap/RealSwapService.kt
antonijzelinskij, line 1: blocking
This approach is architecturally invalid.

Fee is applied only in swap execution, but not in swap estimation.
Handling hydration swap logic is made right in SwapService that is a bad approach since we have other swap sources that we also have to support. Swap service must be independent from any swap sources implementations.

So to fix this please use AtomicSwapOperation that already implemented by HydraDxOperation and encapsulate all specific logic there.
Supporting this inside will be with a guarantee we will consider fee correctly.

Code in swap function must be not edited in this approach.
To determine last hydration segment please use constructAtomicOperations. But you have remember that we must to keep this logic reusable and isolated of any swap source (like hydration). So you have to watch all operations and apply some swap context that will be used by operation to find last segment if needed

feature-swap-impl/src/main/java/io/novafoundation/nova/feature_swap_impl/domain/swap/RealSwapService.kt
```diff
@@ -282,6 +314,59 @@ internal class RealSwapService(
...
+        operation: AtomicSwapOperation,
+        actualSwapLimit: SwapLimit
+    ): Balance {
+        val chainId = operation.assetOut.chainId
+        val chain = chainRegistry.getChain(chainId)
+
+        if (!chain.swap.hydraDxSupported()) return BigInteger.ZERO
+
+        return NovaSwapCommission.feeAmount(actualSwapLimit.amountOutMin).atLeastZero()
+    }
```
antonijzelinskij: blocking

Must be considered in HydraDxOperation

feature-swap-impl/src/main/java/io/novafoundation/nova/feature_swap_impl/domain/swap/RealSwapService.kt
```diff
@@ -282,6 +314,59 @@ internal class RealSwapService(
...
+
+    /**
+     * Builds commission transfer actions to bundle with the swap extrinsic on Hydration.
+     * Returns null if the operation is not on a Hydration chain.
+     *
+     * The commission is [NovaSwapCommission.FEE_PERCENT_DISPLAY]% of the swap output,
+     * calculated from [SwapLimit.amountOutMin] (the post-slippage minimum) for safety.
+     */
+    private suspend fun buildCommissionActions(
```
antonijzelinskij: blocking

Must be cobsidered in HydraDxOperation

feature-swap-impl/src/main/java/io/novafoundation/nova/feature_swap_impl/domain/swap/RealSwapService.kt
```diff
@@ -264,6 +283,17 @@ internal class RealSwapService(
...
     }
 
+    /**
+     * Finds the index of the last [HydraDxOperation] segment, or -1 if none.
+     *
+     * Operation-type detection is required: chain-id detection over-matches
+     * cross-chain transfer segments landing on Hydra, which silently drop
+     * bundleExtraActions and would skip commission on-chain.
+     */
+    private fun findLastHydraSegmentIndex(segments: List<SwapFee.SwapSegment>): Int {
```
antonijzelinskij: Let's not use hydra or any other swap sources here. We must to keep logic in swap service isolated of any sources

feature-swap-impl/src/main/java/io/novafoundation/nova/feature_swap_impl/domain/swap/RealSwapService.kt
```diff
@@ -272,7 +302,9 @@ internal class RealSwapService(
 
         val segmentSubmissionArgs = AtomicSwapOperationSubmissionArgs(actualSwapLimit)
 
-        return operation.submit(segmentSubmissionArgs)
+        val commissionActions = buildCommissionActions(operation, actualSwapLimit)
+
+        return operation.submit(segmentSubmissionArgs, commissionActions)
```
antonijzelinskij: As I see you applied swap fee for any swap operation even if it's not hydration.
This is appeared because of bad architecture approach, but after refactoring SwapService logic and basing on HydraDxOperation we don't have this anyway since all logic will be already applied on HydraDxOperation side

feature-swap-impl/src/main/java/io/novafoundation/nova/feature_swap_impl/domain/swap/RealSwapService.kt
```diff
@@ -391,14 +476,30 @@ internal class RealSwapService(
...
+        // (UI, validations, rate display) sees what the user actually receives. The
+        // gross pool quote remains accessible via quotedPath.lastSegmentQuote for
+        // anything that needs it (notably segment-level SwapLimit construction, which
+        // drives the on-chain extrinsic).
+        val involvesHydraSwap = quotedTrade.path.any { it.edge is HydraDxQuotableEdge }
+        val amountOutNet = if (involvesHydraSwap) {
+            NovaSwapCommission.amountOutAfterFee(amountOutGross)
+        } else {
+            amountOutGross
+        }
```
antonijzelinskij: This code have to be reverted. We will apply this fee on Operation side after refactorign

feature-swap-impl/src/main/java/io/novafoundation/nova/feature_swap_impl/presentation/common/details/SwapConfirmationDetailsFormatter.kt
```diff
@@ -47,12 +47,14 @@ class RealSwapConfirmationDetailsFormatter(
...
 
+        val displayRate = quote.swapRate()
+
         return SwapConfirmationDetailsModel(
             assets = SwapAssetsView.Model(
                 assetIn = formatAssetDetails(chainIn, assetIn, quote.planksIn),
                 assetOut = formatAssetDetails(chainOut, assetOut, quote.planksOut)
             ),
-            rate = formatRate(quote.swapRate(), assetIn, assetOut),
+            rate = formatRate(displayRate, assetIn, assetOut),
```
antonijzelinskij: Not necessary to refactor this)

feature-swap-impl/src/main/java/io/novafoundation/nova/feature_swap_impl/presentation/common/navigation/SwapFlowScope.kt
```diff
@@ -21,21 +20,15 @@ class RealSwapFlowScopeAggregator : SwapFlowScopeAggregator {
...
             synchronized(lock) {
                 scopes -= screenScope
 
                 if (scopes.isEmpty()) {
-                    Log.d("Swaps", "Last swap screen scope was cancelled, cancelling flow scope")
-
                     aggregatedScope!!.cancel()
                     aggregatedScope = null
-                } else {
-                    Log.d("Swaps", "Swap screen scope was cancelled, remaining count: ${scopes.size}")
```
antonijzelinskij, line 38: Let's keep logs please

feature-swap-impl/src/main/java/io/novafoundation/nova/feature_swap_impl/presentation/confirmation/SwapConfirmationFragment.kt
```diff
@@ -64,5 +65,8 @@ class SwapConfirmationFragment : BaseFragment<SwapConfirmationViewModel, Fragmen
         viewModel.slippageAlertMixin.slippageAlertMessage.observe { binder.swapConfirmationAlert.setMessageOrHide(it) }
 
         viewModel.validationInProgress.observe(binder.swapConfirmationButton::setProgressState)
+
+        binder.swapNovaFeeDisclaimer.text = viewModel.novaFeeDisclaimerText
+        viewModel.showNovaFeeDisclaimer.observe { binder.swapNovaFeeDisclaimer.setVisible(it) }
```
antonijzelinskij: Let's set disclamer text in initViews(). subscribe() fun is only for observing view model state

feature-swap-impl/src/main/java/io/novafoundation/nova/feature_swap_impl/presentation/execution/SwapExecutionViewModel.kt
```diff
@@ -86,8 +88,9 @@ class SwapExecutionViewModel(
         executeSwap()
     }
 
-    fun rateClicked() {
-        launchSwapRateDescription()
+    fun rateClicked() = launchUnit {
+        val includesFee = swapStateFlow.first().quote.involvesHydraSwap()
+        launchSwapRateDescription(resourceManager, if (includesFee) NovaSwapCommission.FEE_PERCENT_DISPLAY else null)
```
antonijzelinskij: Duplicated code

feature-swap-impl/src/main/java/io/novafoundation/nova/feature_swap_impl/presentation/confirmation/SwapConfirmationViewModel.kt
```diff
@@ -173,7 +186,11 @@ class SwapConfirmationViewModel(
     }
 
     fun rateClicked() {
-        launchSwapRateDescription()
+        launch {
+            val state = confirmationStateFlow.first()
+            val includesFee = state.swapQuote.involvesHydraSwap()
+            launchSwapRateDescription(resourceManager, if (includesFee) NovaSwapCommission.FEE_PERCENT_DISPLAY else null)
```
antonijzelinskij: Duplicated code

feature-swap-impl/src/main/java/io/novafoundation/nova/feature_swap_impl/presentation/main/SwapMainSettingsFragment.kt
```diff
@@ -96,5 +96,8 @@ class SwapMainSettingsFragment : BaseFragment<SwapMainSettingsViewModel, Fragmen
         }
 
         viewModel.validationProgress.observe(binder.swapMainSettingsContinue::setProgressState)
+
+        binder.swapNovaFeeDisclaimer.text = viewModel.novaFeeDisclaimerText
```
antonijzelinskij: Let's put it to initViews also

feature-swap-impl/src/main/java/io/novafoundation/nova/feature_swap_impl/presentation/main/SwapMainSettingsViewModel.kt
```diff
@@ -239,6 +241,22 @@ class SwapMainSettingsViewModel(
         .distinctUntilChanged()
         .shareInBackground()
 
+    val showNovaFeeDisclaimer: Flow<Boolean> = quotingState.mapNotNull {
+        when (it) {
+            is QuotingState.Loaded -> it.quote.involvesHydraSwap()
+            is QuotingState.Default,
+            is QuotingState.Error -> false
+            else -> null
```
antonijzelinskij: Sugges to not use else in when. Let's write Loading here.
In this approach we will be notified by compiler if we added new type

feature-swap-impl/src/main/java/io/novafoundation/nova/feature_swap_impl/presentation/main/SwapMainSettingsViewModel.kt
```diff
@@ -350,7 +368,11 @@ class SwapMainSettingsViewModel(
     }
 
     fun rateDetailsClicked() {
-        launchSwapRateDescription()
+        launch {
+            val includesFee = (quotingState.value as? QuotingState.Loaded)
+                ?.quote?.involvesHydraSwap() ?: false
+            launchSwapRateDescription(resourceManager, if (includesFee) NovaSwapCommission.FEE_PERCENT_DISPLAY else null)
```
antonijzelinskij: Duplicated code

antonijzelinskij (changes requested):

Leemo94:
**Known limitation surfaced during QA (pre-existing, not introduced by this PR)**

Small cross-chain inbound swap amounts (e.g. Asset Hub DOT → swap-on-Hydration at 0.01 DOT input) fail at the cross-chain dry-run stage. The chain-side cause is that the XCM execution on the destination doesn't emit a deposit event the dry-runner can detect — `RealXcmTransferDryRunner.searchDepositAmount` then throws `IllegalStateException("No deposits detected")`, which is caught and remapped to the generic `SwapOperationSubmissionException.SimulationFailed` at `CrossChainTransferAssetExchange.dryRunTransfer:299`, so the user sees a developer-only "Simulation of swap step failed" with no remediation hint.

The same failure reproduces on iOS production with the same DOT input amount, confirming it predates this PR (verified locally that 0.1 DOT cleared the gate while 0.01 DOT did not).

Out of scope here, but a follow-up could thread the underlying `Throwable` through `SimulationFailed`, introduce a typed variant (e.g. `AmountBelowDestinationMinimum`), and let the UI render something like _"Amount too small for this cross-chain route — try a larger amount"_.

feature-swap-impl/src/main/java/io/novafoundation/nova/feature_swap_impl/domain/swap/RealSwapService.kt
Leemo94: Fully restructured in 207b4429c.

Summary of the changes:
- `RealSwapService.swap()` / `submitFirstSwapStep` / `quoteInternal` no longer reference `HydraDxOperation`, `HydraDxQuotableEdge`, or commission math. `findLastHydraSegmentIndex`, `getCommissionAmount`, `buildCommissionActions` are gone.
- `AtomicSwapOperationSubmissionArgs` gains a generic `isLastSwapOperation` flag; `AtomicSwapOperation.execute/submit` lose the `bundleExtraActions` parameter (no longer needed as an API surface). The `BundleExtraActions` typealias is removed.
- `AtomicSwapOperationPrototype` gains a default-identity `postProcessFinalAmountOut(amountOut)` hook. `quoteInternal` asks the final prototype to post-process the displayed amountOut.
- `HydraDxOperation` now owns commission internally: appends the transfer call inside `submitInternal` when `args.isLastSwapOperation`, and returns `actualReceivedAmount` net of commission from `execute()`. `HydraDxOperationPrototype.postProcessFinalAmountOut` overrides to deduct 0.85% so the displayed quote is net.
- One deliberate deviation: commission is NOT added to `estimateFee()` (only to submission). A first cut included it, but the extra fee weight propagates via `convertIntermediateSegmentsFeesToAssetIn` → `additionalAmountForSwap` → segment 0's transfer amount, which breaks cross-chain inbound swaps at dry-run. Marginal estimation inaccuracy (~1 transfer call's weight) sits comfortably inside the existing `FEE_QUOTE_BUFFER = 1.1` headroom.
- iOS PR #1824 still follows the previous layering; will mirror when iOS reviewer weighs in.

feature-swap-impl/src/main/java/io/novafoundation/nova/feature_swap_impl/domain/swap/RealSwapService.kt
Leemo94: Done in 207b4429c — collapsed into `HydraDxOperation.novaCommissionAmount(actualSwapLimit)`. See the umbrella thread #3276088371 for the full picture.

feature-swap-impl/src/main/java/io/novafoundation/nova/feature_swap_impl/domain/swap/RealSwapService.kt
Leemo94: Done in 207b4429c — moved to `HydraDxOperation.appendNovaCommissionCall`, invoked from `submitInternal` when `args.isLastSwapOperation`. RealSwapService no longer constructs any transfer calls. See the umbrella thread #3276088371.

feature-swap-impl/src/main/java/io/novafoundation/nova/feature_swap_impl/domain/swap/RealSwapService.kt
Leemo94: Done in 207b4429c — function removed. RealSwapService now passes a generic `isLastSwapOperation = (index == segments.lastIndex)` flag in `AtomicSwapOperationSubmissionArgs`; only `HydraDxOperation` consults it. No source-specific code remains in the service layer.

feature-swap-impl/src/main/java/io/novafoundation/nova/feature_swap_impl/domain/swap/RealSwapService.kt
Leemo94: Correct — that fell out of the restructure in 207b4429c. `operation.execute(args)` no longer subtracts commission centrally; each operation decides for itself. Cross-chain and asset-conversion operations ignore the flag; only `HydraDxOperation` charges.

feature-swap-impl/src/main/java/io/novafoundation/nova/feature_swap_impl/domain/swap/RealSwapService.kt
Leemo94: Reverted in 207b4429c. `RealSwapService.quoteInternal` no longer has the `involvesHydraSwap` / `amountOutNet` branch — it asks `prototypes.last().postProcessFinalAmountOut(amountOutGross)`, and `HydraDxOperationPrototype` overrides to deduct commission. The priceImpact still uses the gross pool quote, since computing it against the post-commission amount inflates the displayed slippage by ~0.85pp (separate concern, kept the original behaviour).

common/src/main/res/values-fr-rFR/strings.xml
Leemo94: Done in 55cadf425. `swap_rate_includes_fee_description` removed from all `values-*` locales; only `values/strings.xml` (English) remains.

feature-assets/src/main/java/io/novafoundation/nova/feature_assets/data/repository/TransactionHistoryRepository.kt
Leemo94: Done in 55cadf425. Introduced `HydrationSwapTransferFilterFactory` (interface in `feature-swap-api/data/history/`) + `RealHydrationSwapTransferFilterFactory` (impl in `feature-swap-impl/data/history/`), wired via `SwapFeatureModule` / `SwapFeatureApi` and consumed in `AssetsFeatureModule.provideTransactionHistoryRepository`. `RealTransactionHistoryRepository.createTransactionFilters` now just calls `hydrationSwapTransferFilterFactory.create(chain)`.

feature-assets/src/main/java/io/novafoundation/nova/feature_assets/data/repository/TransactionHistoryRepository.kt
Leemo94: Done in 55cadf425. `IgnoreTransfersInvolvingSystemAccount` moved into `RealHydrationSwapTransferFilterFactory.kt` as a private class. RealTransactionHistoryRepository no longer carries that filter class.

feature-swap-api/src/main/java/io/novafoundation/nova/feature_swap_api/presentation/view/bottomSheet/description/DescriptionBottomSheetLauncherExt.kt
Leemo94: Done in 55cadf425. Added `sealed class SwapRateDescriptionMode { object Default ; data class IncludesFee(val feePercentDisplay: String) }`. `launchSwapRateDescription(resourceManager, mode)` takes it directly. Named the class `SwapRateDescriptionMode` rather than `FeeDisplay` to avoid colliding with the existing `feature_wallet_api.presentation.mixin.fee.model.FeeDisplay`.

feature-swap-impl/src/main/java/io/novafoundation/nova/feature_swap_impl/presentation/common/details/SwapConfirmationDetailsFormatter.kt
Leemo94: Reverted in 55cadf425 — inlined back to `formatRate(quote.swapRate(), assetIn, assetOut)`.

feature-swap-impl/src/main/java/io/novafoundation/nova/feature_swap_impl/presentation/common/navigation/SwapFlowScope.kt
Leemo94: Restored in 55cadf425 — all three `Log.d("Swaps", ...)` lines in `RealSwapFlowScopeAggregator` are back.

feature-swap-impl/src/main/java/io/novafoundation/nova/feature_swap_impl/presentation/confirmation/SwapConfirmationFragment.kt
Leemo94: Done in 55cadf425 — `binder.swapNovaFeeDisclaimer.text = viewModel.novaFeeDisclaimerText` moved into `initViews()`; the visibility observer stays in `subscribe()`.

feature-swap-impl/src/main/java/io/novafoundation/nova/feature_swap_impl/presentation/execution/SwapExecutionViewModel.kt
Leemo94: Done in 55cadf425. Extracted `SwapQuote.swapRateDescriptionMode()` helper in `SwapQuoteNovaFee.kt`; all three call sites (`SwapExecutionViewModel.rateClicked`, `SwapConfirmationViewModel.rateClicked`, `SwapMainSettingsViewModel.rateDetailsClicked`) collapse to `launchSwapRateDescription(resourceManager, quote.swapRateDescriptionMode())`.

feature-swap-impl/src/main/java/io/novafoundation/nova/feature_swap_impl/presentation/confirmation/SwapConfirmationViewModel.kt
Leemo94: Done in 55cadf425 via the same `SwapQuote.swapRateDescriptionMode()` helper — see #3276438909.

feature-swap-impl/src/main/java/io/novafoundation/nova/feature_swap_impl/presentation/main/SwapMainSettingsFragment.kt
Leemo94: Done in 55cadf425 — same move as #3276395516.

feature-swap-impl/src/main/java/io/novafoundation/nova/feature_swap_impl/presentation/main/SwapMainSettingsViewModel.kt
Leemo94: Done in 55cadf425 — `else -> null` is now `is QuotingState.Loading -> null`.

feature-swap-impl/src/main/java/io/novafoundation/nova/feature_swap_impl/presentation/main/SwapMainSettingsViewModel.kt
Leemo94: Done in 55cadf425 via the same `SwapQuote.swapRateDescriptionMode()` helper — see #3276438909.

PR #2277, opened 2026-04-10
Add mandatory consent banner for wallet creation and existing-user upgrade

Leemo94:
> ⚠️ **Requires approved ToS & Privacy Notice + URLs of ToS & Privacy Notice on novasama.io before merging**

## Purpose

Adds a mandatory ToS / Privacy Notice consent mechanism. Users must
affirmatively accept before creating or connecting a wallet. Existing
users are prompted on first launch after the update.

Companion iOS PR: novasamatech/nova-wallet-ios#1826

### New-user flow (Welcome screen)
- Consent checkbox with required text placed above the Create / Already
  Have buttons
- Both buttons disabled until the checkbox is checked via LiveData
  observer in WelcomeFragment
- "Terms of Service" and "Privacy Notice" are underlined clickable spans
  that open the system browser (Intent.ACTION_VIEW)
- Acceptance persisted via `ConsentRepository` in the `common` module,
  backed by SharedPreferences with a versioned integer key

### Existing-user upgrade flow
- `RootViewModel.checkForConsentBannerUpgrade()` checks consent state
  on launch
- Gated behind `AutomaticInteractionGate.awaitInteractionAllowed()` so
  the dialog appears after PIN entry, not over the PIN screen
- `ConsentBannerUpgradeDialogFragment` with `setCancelable(false)` and
  `setCanceledOnTouchOutside(false)` — blocks back press and tap-outside
- On accept, writes consent via `RootViewModel.consentBannerUpgradeAccepted()`

### Dagger wiring
- `ConsentRepository` provided at `@ApplicationScope` in `CommonModule`,
  exposed via `CommonApi.consentRepository()`
- Accessible from both `WelcomeModule` (onboarding) and
  `RootActivityModule` (app root) without cross-module leakage
- `OnboardingFeatureDependencies` and `RootDependencies` updated to
  expose `ConsentRepository` through the dependency chain

### Privacy compliance
- Consent state is **device-local only** (SharedPreferences). No
  PostHog, no Firebase Analytics, no server-side log.
- Placeholder URLs (`TODO_CONSENT_URL_*`) must be replaced with the
  canonical novasama.io pages before merging (dependent on rollout
  action items C-01 / C-02)

🤖 Generated with [Claude Code](https://claude.com/claude-code)


common/src/main/java/io/novafoundation/nova/common/data/preferences/ConsentBannerConstants.kt
```diff
@@ -0,0 +1,31 @@
...
+     * must be re-collected. The persisted accepted version is compared against this value
+     * to decide whether the user has accepted the current revision.
+     */
+    const val CURRENT_CONSENT_VERSION: Int = 1
+
+    /**
+     * Placeholder until rollout action item C-01 lands the canonical novasama.io page.
+     * MUST be replaced before merging to the upstream develop branch.
+     */
+    const val TERMS_OF_SERVICE_URL: String = "https://novasama.io/TODO_CONSENT_URL_TERMS"
```
valentunn, line 24: Pending TODO here and in `PRIVACY_NOTICE_URL`

app/src/main/java/io/novafoundation/nova/app/root/presentation/consentBanner/ConsentBannerUpgradeDialogFragment.kt
```diff
@@ -0,0 +1,115 @@
...
+        super.onViewCreated(view, savedInstanceState)
+
+        // isCancelable on the DialogFragment also blocks the back press.
+        isCancelable = false
+
+        val titleTextView = view.findViewById<TextView>(R.id.consentUpgradeTitle)
+        val descriptionTextView = view.findViewById<TextView>(R.id.consentUpgradeDescription)
+        val consentTextView = view.findViewById<TextView>(R.id.consentUpgradeText)
+        val checkBox = view.findViewById<CheckBox>(R.id.consentUpgradeCheckbox)
+        val acceptButton = view.findViewById<PrimaryButton>(R.id.consentUpgradeAccept)
```
valentunn: We use data-binding to access views. Extend `BaseBottomSheetFragment` and use `binder.<view>` to access views

app/src/main/java/io/novafoundation/nova/app/root/presentation/consentBanner/ConsentBannerUpgradeDialogFragment.kt
```diff
@@ -0,0 +1,115 @@
...
+    companion object {
+        const val TAG = "ConsentBannerUpgradeDialogFragment"
+    }
+
+    /**
+     * Set by RootActivity before showing the dialog. Invoked when the user
+     * checks the box and taps Accept. RootActivity persists the acceptance
+     * via RootViewModel.consentBannerUpgradeAccepted().
+     */
+    var onAccepted: (() -> Unit)? = null
```
valentunn: The fragment should have an assicoated ViewModel that holds the screen state (check-box checked, button enabled via `DescriptiveButtonState`) and performs navigation via `RootRouter.back`

Also I think this screen should live in onboarding feature module

app/src/main/java/io/novafoundation/nova/app/root/presentation/consentBanner/ConsentBannerUpgradeDialogFragment.kt
```diff
@@ -0,0 +1,115 @@
...
+    override fun onResume() {
+        super.onResume()
+
+        // Default Android dialog width is theme-dependent and typically much
+        // narrower than the screen, which causes the consent text to wrap
+        // awkwardly. Stretch the dialog window to ~92% of the screen width
+        // so the consent paragraph reads naturally.
+        val window = dialog?.window ?: return
+        val widthPixels = (resources.displayMetrics.widthPixels * 0.92f).toInt()
+        window.setLayout(widthPixels, WindowManager.LayoutParams.WRAP_CONTENT)
```
valentunn: We do not use dialogs for non trivial screens. I think we should switch to bottom sheet via `BaseBottomSheetFragment`

valentunn (commented):
Related to this and all other PRs: can you please attach screenshots of working feature?

common/src/main/java/io/novafoundation/nova/common/data/preferences/ConsentBannerConstants.kt
Leemo94: These are intentional placeholders — the canonical novasama.io pages depend on rollout items C-01 (ToS) and C-02 (Privacy Notice). The strings are syntactically valid URLs so nothing crashes; they resolve to a 404. The "MUST be replaced before merging" KDoc on each constant records the gate, and merge to develop is blocked behind those rollout items per the warning in the PR body. Happy to add a debug-build `check()` in the object's init block as an extra safeguard if useful.

app/src/main/java/io/novafoundation/nova/app/root/presentation/consentBanner/ConsentBannerUpgradeDialogFragment.kt
Leemo94: Switched to `BaseBottomSheetFragment` with a dedicated `ConsentBannerUpgradeViewModel` (in `feature-onboarding-impl`) that exposes `acceptButtonState: Flow<DescriptiveButtonState>` and is bound via `binder.consentUpgradeAccept.setState(...)` — should match the pattern you had in mind.

One small deviation worth flagging: `RootRouter` has no `back()` method (only specific destination openers), so I initially routed through `OnboardingRouter.back()` → `BaseNavigator.back()` → `firstAttachedHolder.executeBack()`. During QA that resolved to the split-screen nav controller (the wallet UI behind the dialog), not the `DialogFragmentNavigator` managing this `<dialog>` destination — `popBackStack()` on the wrong controller fell through to `activity.finish()`, closing the app on Accept. Replaced with a `dismissEvent: LiveData<Event<Unit>>` fired by the ViewModel and observed by the Fragment, which calls `DialogFragment.dismiss()` directly. The ViewModel still owns the dismissal intent (Fragment is a passive observer) and `DialogFragmentNavigator` cleans up the right back-stack entry. Happy to swap to a dedicated `RootRouter.closeConsentBannerUpgrade()` that pops the root nav holder explicitly if you'd prefer that shape.

Leemo94:
Adds unskippable consent popup. Users (new and existing) must accept prior to using nova wallet.

returning 1
returning 2
new 2
new 1

PR #2286, opened 2026-07-02
Pin scaleway/action-scw-secret to last working commit

stepanLav:
## Problem

CI workflows that fetch secrets via `scaleway/action-scw-secret` started failing on 2026-07-02 with:

```
Failed to fetch secret: '...'. Error: PermissionsDeniedError: insufficient permissions: list secret.
```

Nothing changed on our side — GitHub secrets, the Scaleway IAM policy, and the API key are all intact. The trigger is upstream: scaleway/action-scw-secret#145 was merged on 2026-07-02 at 07:13 UTC and the floating `v0` tag picked it up. The new version passes `organization_id` to the `ListSecrets` call, which requires organization-wide list permission. Our CI keys are intentionally scoped to a single project, so every secret fetch now fails.

## Fix

Pin the action to `9f2ead1e62c0b201b34035cd2ae5274f64c58787` — the last release before the regression. Pinning by SHA also protects us from similar surprises in the future.

Reported upstream: scaleway/action-scw-secret#146.

### DIFF
--- a/.github/actions/install/action.yml
+++ b/.github/actions/install/action.yml
@@ -23,7 +23,7 @@ runs:
   using: "composite"
   steps:
     - name: 🔐 Fetch secrets from Scaleway
-      uses: scaleway/action-scw-secret@v0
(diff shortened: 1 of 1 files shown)


stepanLav:
problem fixed

PR #2290, opened 2026-07-10
Fixed nominator payouts

antonijzelinskij:
image


ERussel (approved):

feature-staking-impl/src/main/java/io/novafoundation/nova/feature_staking_impl/data/repository/PayoutRepository.kt
```diff
@@ -86,6 +84,16 @@ class PayoutRepository(
         }
     }
 
+    private fun resolveNominatorPayoutTargets(
+        stakingState: StakingState.Stash.Nominator,
+        historicalRange: List<BigInteger>,
+    ): List<PayoutTarget> {
+        val currentTargets = stakingState.nominations.targets.map { it.intoKey() }
```
ERussel, line 91: That prevents displaying previous payouts if a user changed validators

ERussel (commented):
Accidentally approved but solution looks incorrect since it doesn't account validators change. We need to research why subquery is missing validators data

PR #2293, opened 2026-07-15
Fix DOT staking APY: derive from the staking pallet's era reward allocation

### LINKED ISSUE #2292: DOT staking APY overstated ~3x (shows ~8.9%, actual ~2.93%) — reward calc ignores DAP staker split
## Summary
The Polkadot (DOT) staking screen shows an APY of **~8.9%** (max validator; ~6.5% average) when the actual staking return is **~2.93%** — overstated by ~3×.

## Root cause
Since the Dynamic Allocation Pool (DAP) reform, DOT staking rewards are a fixed allocation of the period mint, not the whole mint. Of the ~153,132 DOT minted per day, only **45.2% (~69,216 DOT/day) is paid to stakers**; the rest goes to the validator self-stake incentive (22.6%) and the DAP buffer (32.2%).

The reward calc reads the `Inflation.experimental_issuance_prediction_info` runtime API and binds `next_mint.0` as `toStakers` — but that value is the **full** period mint (~153,132 DOT), not the staker allocation. `calculateStakersInflation` then divides the full mint by total issuance, overstating the pool by **2.21× (= 1 / 0.452)**; the headline applies the max-validator factor (~1.38×), landing at ~8.9%.

- `feature-staking-api/src/main/java/io/novafoundation/nova/feature_staking_api/domain/model/InflationPredictionInfo.kt:35` — `toStakers = bindNumber(toStakersRaw)` (takes `list[0]`, the full mint)
- `feature-staking-impl/src/main/java/io/novafoundation/nova/feature_staking_impl/domain/rewards/RewardCalculatorFactory.kt:104` — routing: `customRelayChainCalculator` → `POLKADOT_ASSET_HUB` → `PolkadotInflationPrediction` → `InflationPredictionInfoCalculator`. (The pre-DAP `RewardCurveInflationRewardCalculator` is **not** on the DOT path.)

## Evidence (on-chain, Polkadot Asset Hub)
- DAP mint: 127.61 DOT / 72s = 153,132 DOT/day (`dap.IssuanceMinted`)
- `Dap.BudgetAllocation`: staker_rewards **45.2%**, validator_incentive 22.6%, buffer 32.2%
- `Staking.ErasValidatorReward` ≈ **69,216 DOT/era** (= 45.2% of the daily mint)
- `Staking.ErasTotalStake` ≈ 862,000,000 DOT
- Actual APR = 69,216 × 365 ÷ 862,000,000 = **2.93%**

## Impact
Every user browsing DOT staking sees ~9% when they will actually earn ~3% — an inflated-yield display.

## Suggested fix
Drive the APY from **`Staking.ErasValidatorReward`** (the realised per-era staker payout) instead of `next_mint.toStakers`. The return reduces to `ErasValidatorReward × erasPerYear / ErasTotalStake` = the correct ~2.93%. No hardcoded split, and it auto-tracks both the issuance curve (ref 1710) and any governance re-allocation of the DAP budget.

## Notes
- Nomination-pool APY inherits this automatically (`RealNominationPoolRewardCalculator` wraps the relaychain calculator).
- Kusama and other relaychains are unaffected (they keep the inflation-curve path).
- Same bug on iOS → novasamatech/nova-wallet-ios#1836.

Leemo94:
Closes #2292

## Problem

The Polkadot staking screens show ~8.9% APY (max validator) while the actual staking return is ~2.93%. Since the DAP reform, only the staker allocation of the period mint (currently 45.2%) is paid to stakers, but the reward calculator treated the `Inflation` runtime api's full `next_mint` (~153,132 DOT/day) as the staker reward — a 2.21× (= 1/0.452) overstatement, further amplified by the max-validator headline.

The root cause is upstream: fellows-runtimes v2.3.0 (2026-06-05) changed `next_mint`'s meaning on Polkadot from `(stakers, treasury)` to `(total emission, 0)` without changing the type — the split now happens downstream in pallet-dap. The same field still means "staker amount" on Kusama, so no interpretation of `next_mint.0` alone can be correct on both chains.

## Fix

Query the staking pallet's own view function instead:

```
Staking::era_reward_allocation(activeEra − 1) -> { staker_rewards: u128, validator_incentive: u128 }
```

`staker_rewards` is the exact per-era staker budget payouts draw from (mirrors `Staking.ErasValidatorReward`, recorded at era end — hence `activeEra − 1`, which the route's existing `EraTimeCalculator` already provides). The calculator keeps the `InflationBasedRewardCalculator` contract, so the per-validator / commission / compound machinery is untouched. No hardcoded split: the value tracks the ref-1710 issuance curve and any governance re-allocation of the DAP budget automatically. Verified live on both networks: Polkadot era 2229 → 69,215.6368 DOT → 2.93%; on Kusama the same call returns the correct legacy-mode figures (unchanged ~15.45%), and it also returns correct values for pre-DAP eras — one semantic across reward regimes.

The call is a single `state_call` to `RuntimeViewFunction_execute_view_function` via the existing `SocketService.stateCall` helper. View functions are not declared as runtime apis in the metadata version the app requests (their ids/types ship with v16), so the request is assembled statically — the id is derived the same way storage keys are:

```
twox128("Staking") ++ twox128("era_reward_allocation(EraIndex) -> crate::reward::EraRewardAllocation<BalanceOf<T>>")
```

and the response decoded with static SCALE coding. A unit test pins the derived id byte-for-byte to the value served in on-chain metadata.

## Changes

- `EraRewardAllocation` (new model): id derivation + static decode of the `Result<Vec<u8>, ViewFunctionDispatchError>` envelope and the allocation payload.
- `StakingRepository.getEraRewardAllocation` replaces `getInflationPredictionInfo`; `StakingRepositoryImpl` assembles the state call (`useScaleWriter` + `StateCallRequest`), and its now-unused `MultiChainRuntimeCallsApi` dependency is dropped.
- `EraRewardAllocationRewardCalculator` replaces `InflationPredictionInfoCalculator` — same base-class contract, yearly inflation expressed against total issuance so the staked-portion division cancels to `stakerRewards × erasPerYear ÷ totalStaked`.
- `RewardCalculatorFactory`: the `POLKADOT_ASSET_HUB` route builds the new calculator inside the existing `runCatching`; a zero allocation is treated as an error so the existing fallback applies rather than rendering 0%.
- Nomination-pools APY inherits the fix automatically (the pools calculator delegates to this route).
- Tests: view-function id pin, decoder vectors captured verbatim from mainnet (success, zero, error envelope, truncation), and ~2.93% return assertions on current on-chain inputs.

## Notes for reviewers

- `validator_incentive` (the DAP 22.6% pot) is intentionally **not** part of the displayed return — it is paid only to validator stashes, weighted by self-stake.
- Kusama and other chains are untouched (they keep their existing calculators); Kusama could be routed through the same view function in a follow-up to retire the per-chain `next_mint` semantics divergence for good (verified working).
- The dashboard "estimated earnings" tile comes from the backend multi-staking API, not this calculator — if the backend derives its rate the same way, it needs the equivalent fix server-side.

Same fix on iOS: novasamatech/nova-wallet-ios#1837
Backend leg (dashboard cards / `stakingApies`): novasamatech/subquery-staking#104

🤖 Generated with [Claude Code](https://claude.com/claude-code)


valentunn (approved):
Ok for hotfix but we will need to clean up to use view functions from runtime metadata later

PR #2314, opened 2026-08-14
Fix/hydra commission estimation

antonijzelinskij:
## Problem

Swaps on Hydration that pay the fee in a non-native asset were failing with
`Token: FundsUnavailable`. Max-amount swaps hit it every time, and each failed
attempt still burned the fee (~2.7 MYTH).

The cause was fee under-estimation. To convert the native fee into the fee asset
we asked the router for the cheapest swap path. But the chain never swaps the
fee — it converts it with a 10-minute EMA oracle price along the route registered
in `Router.Routes`. Taking the cheapest of several candidate paths systematically
lands on whichever pool is furthest from the market, so our estimate came out too
low. For the failing swap we showed 1.85 MYTH while the chain charged 2.69.

## Fix

New `HydrationOraclePriceConverter` that mirrors what `pallet-transaction-multi-payment`
actually does:

- takes the route from `Router.Routes`, with the same ordered-pair lookup,
  inversion, and single-Omnipool-hop default when no route is registered
- prices each hop the way the runtime does — Omnipool through the LRNA hub,
  Stableswap through the pool's share token, XYK directly, Aave 1:1
- advances stale oracle entries to the current block like
  `EmaOracle::get_updated_entry`, instead of using the stored value as-is
- finishes with the same fixed-point steps as `convert_fee_with_price`

Asset ids (native, hub) come from runtime constants. Pool names, oracle source ids
and the EMA smoothing factor are collected in `HydrationOnChain`, each with a link
to its definition in the Hydration runtime.

The fallback order now matches the chain as well: oracle price → `AcceptedCurrencies`.
A best-path quote is kept only as a last resort if both are unavailable.

The same bug existed in `HydrationConversionFeePayment`, which handles fees for
non-swap transactions (transfers, XCM out of Hydration) when paying in a custom
asset. Fixed there too.

## Verification

Checked against two real failed transactions (Hydration blocks 13596237 and
13596264) by comparing with the fee the chain actually withdrew. The new
conversion reproduces it exactly, down to the last plank. Reading oracle entries
without the fast-forward was off by a small margin; the old quote-based path was
off by ~30%.

Swap also tested on device.

### DIFF
--- a/feature-swap-core/src/main/java/io/novafoundation/nova/feature_swap_core/data/assetExchange/conversion/types/hydra/HydrationOracleApi.kt
+++ b/feature-swap-core/src/main/java/io/novafoundation/nova/feature_swap_core/data/assetExchange/conversion/types/hydra/HydrationOracleApi.kt
@@ -1,0 +1,122 @@
+@file:Suppress("RedundantUnitExpression")
+
+package io.novafoundation.nova.feature_swap_core.data.assetExchange.conversion.types.hydra
+
+import io.novafoundation.nova.common.data.network.runtime.binding.bindNumber
+import io.novafoundation.nova.common.data.network.runtime.binding.castToDictEnum
+import io.novafoundation.nova.common.data.network.runtime.binding.castToList
+import io.novafoundation.nova.common.data.network.runtime.binding.castToStruct
+import io.novafoundation.nova.common.utils.structOf
+import io.novafoundation.nova.common.utils.system
+import io.novafoundation.nova.feature_swap_core_api.data.network.HydraDxAssetId
+import io.novafoundation.nova.runtime.storage.source.query.StorageQueryContext
+import io.novafoundation.nova.runtime.storage.source.query.api.QueryableModule
+import io.novafoundation.nova.runtime.storage.source.query.api.QueryableStorageEntry0
+import io.novafoundation.nova.runtime.storage.source.query.api.QueryableStorageEntry1
+import io.novafoundation.nova.runtime.storage.source.query.api.QueryableStorageEntry3
+import io.novafoundation.nova.runtime.storage.source.query.api.storage0
+import io.novafoundation.nova.runtime.storage.source.query.api.storage1
+import io.novafoundation.nova.runtime.storage.source.query.api.storage3
+import io.novasama.substrate_sdk_android.runtime.definitions.types.composite.DictEnum
+import io.novasama.substrate_sdk_android.runtime.definitions.types.composite.Struct
+import io.novasama.substrate_sdk_android.runtime.metadata.module
+import io.novasama.substrate_sdk_android.runtime.metadata.RuntimeMetadata
+import io.novasama.substrate_sdk_android.runtime.metadata.module.Module
+import java.math.BigInteger
+
+/**
+ * One leg of the route registered on chain. [pool] is the pool type name as the runtime spells it:
+ * `Omnipool`, `XYK`, `Stableswap`, `Aave`. [poolId] is only carried by `Stableswap`, where it is the
+ * pool's share token - the asset its oracle entries are stored against
+ */
+class HydrationRouteHop(
+    val pool: String,
+    val poolId: HydraDxAssetId?,
+    val assetIn: HydraDxAssetId,
+    val assetOut: HydraDxAssetId
+)
+
+/**
+ * A stored `EmaOracle.Oracles` entry: EMA price for an ordered asset pair - planks of the lower asset id
+ * per plank of the higher one - and the block the entry was last updated at
+ */
+class HydrationOracleEntry(
+    val numerator: BigInteger,
+    val denominator: BigInteger,
+    val updatedAt: BigInteger
+)
+
+@JvmInline
+value class HydrationRouterApi(override val module: Module) : QueryableModule
+
+@JvmInline
+value class HydrationEmaOracleApi(override val module: Module) : QueryableModule
+
+@JvmInline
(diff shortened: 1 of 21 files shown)


feature-swap-core/src/main/java/io/novafoundation/nova/feature_swap_core/data/assetExchange/conversion/types/hydra/HydrationOracleApi.kt
```diff
@@ -0,0 +1,122 @@
...
+val RuntimeMetadata.hydrationEmaOracle: HydrationEmaOracleApi
+    get() = HydrationEmaOracleApi(module("EmaOracle"))
+
+context(StorageQueryContext)
+val RuntimeMetadata.hydrationSystem: HydrationSystemApi
+    get() = HydrationSystemApi(system())
+
+context(StorageQueryContext)
+val HydrationSystemApi.number: QueryableStorageEntry0<BigInteger>
+    get() = storage0(name = "Number", binding = { decoded -> bindNumber(decoded) })
```
stepanLav: `runtime` already has this one — `runtime/storage/typed/SystemRuntimeApi.kt` declares `SystemRuntimeApi` with a `number` entry bound through `bindBlockNumber`. Can we drop `HydrationSystemApi` and use `metadata.system.number` in `loadContext` instead? `feature-swap-core` already depends on `runtime`, so it's a straight import. DRY, and it also keeps the pallet-agnostic storage APIs in one place rather than growing a hydra-local copy.

feature-swap-core/src/main/java/io/novafoundation/nova/feature_swap_core/data/assetExchange/conversion/types/hydra/HydrationOracleApi.kt
```diff
@@ -0,0 +1,122 @@
...
+)
+
+@JvmInline
+value class HydrationRouterApi(override val module: Module) : QueryableModule
+
+@JvmInline
+value class HydrationEmaOracleApi(override val module: Module) : QueryableModule
+
+@JvmInline
+value class HydrationSystemApi(override val module: Module) : QueryableModule
```
stepanLav: Three pallets in one file goes against how this package is laid out — `OmnipoolApi.kt`, `XYKApi.kt`, `StableSwapApi.kt`, `TokensApi.kt`, `MultiTransactionPaymentApi.kt` are all one-per-pallet. Once `HydrationSystemApi` is gone (comment above), splitting the rest into `RouterApi.kt` and `EmaOracleApi.kt` would match, with `HydrationRouteHop` / `HydrationOracleEntry` moving next to the API that returns them.

feature-swap-core/src/main/java/io/novafoundation/nova/feature_swap_core/data/assetExchange/conversion/types/hydra/RealHydrationOraclePriceConverter.kt
```diff
@@ -0,0 +1,276 @@
...
+}
+
+private class HydrationFeeContext(
+    val nativeAssetId: HydraDxAssetId,
+    val hubAssetId: HydraDxAssetId,
+    val parentBlock: BigInteger
+)
+
+@FeatureScope
+class RealHydrationOraclePriceConverter @Inject constructor(
```
stepanLav, line 70: Two small deviations from the siblings in this package: `RealHydrationPriceConversionFallback` and `RealHydrationAcceptedFeeCurrenciesFetcher` are both `internal`, and both leave `@FeatureScope` on the `@Binds` method only. Here it sits on the class *and* on `bindHydrationOraclePriceConverter` in `HydraDxBindsModule`, which is redundant. Nothing outside the module uses the concrete type, so:

```suggestion
internal class RealHydrationOraclePriceConverter @Inject constructor(
```

feature-swap-impl/src/main/java/io/novafoundation/nova/feature_swap_impl/data/assetExchange/hydraDx/HydraDxAssetExchange.kt
```diff
@@ -652,9 +656,12 @@ internal class HydraDxAssetExchange(
...
-                .recoverCatching { hydrationPriceConversionFallback.convertNativeAmount(nativeFee.amount, customFeeAsset) }
-                .getOrThrow()
+            // The chain converts the fee with an oracle price and falls back to `AcceptedCurrencies`, never swapping.
+            // A best-path quote lands on whichever pool is furthest from the market, so it is only a last resort here
+            val quotedFee = hydrationOraclePriceConverter.convertNativeAmount(nativeFee.amount, customFeeAsset)
+                ?: runCatching { hydrationPriceConversionFallback.convertNativeAmount(nativeFee.amount, customFeeAsset) }
+                    .recoverCatching { swapHost.quote(args) }
+                    .getOrThrow()
 
             // Fees in non-native assets are especially volatile since conversion happens through swaps so we add some buffer to mitigate volatility
```
stepanLav, line 666: This comment is stale after the change right above it — conversion no longer happens through swaps. The buffer still earns its keep, but for a different reason now (oracle lag + drift between estimation and inclusion), and worth spelling out so nobody deletes it later as obsolete:

```suggestion
            // Oracle prices lag the market - a 10-minute EMA advanced to the current block - and can still move
            // between estimation and inclusion, so we add some buffer to mitigate that
```

stepanLav (approved):

PR #2316, opened 2026-08-18
Implemented staking anouncements

antonijzelinskij:
image image image

### DIFF
--- a/common/src/main/java/io/novafoundation/nova/common/data/announcements/AnnouncementMapper.kt
+++ b/common/src/main/java/io/novafoundation/nova/common/data/announcements/AnnouncementMapper.kt
@@ -15,6 +15,7 @@ private fun mapAnnouncementFromRemote(remote: AnnouncementRemote, languageCode: 
     val description = remote.description.localizedOrDefault(languageCode) ?: return null
 
     return Announcement(
+        chainId = remote.chainId,
         style = mapStyleFromRemote(remote.style),
         description = description
     )
--- a/common/src/main/java/io/novafoundation/nova/common/data/announcements/AnnouncementMapper.kt
+++ b/common/src/main/java/io/novafoundation/nova/common/data/announcements/AnnouncementMapper.kt
@@ -11,8 +11,8 @@ fun mapAnnouncementsFromRemote(
     return remote[section.key].orEmpty().mapNotNull { mapAnnouncementFromRemote(it, languageCode) }
 }
 
-private fun mapAnnouncementFromRemote(remote: AnnouncementRemote, languageCode: String): Announcement? {
-    val description = remote.description.localizedOrDefault(languageCode) ?: return null
+private fun mapAnnouncementFromRemote(remote: AnnouncementRemote?, languageCode: String): Announcement? {
+    val description = remote?.description?.localizedOrDefault(languageCode) ?: return null
 
     return Announcement(
         chainId = remote.chainId,
--- a/common/src/main/java/io/novafoundation/nova/common/data/announcements/AnnouncementMapper.kt
+++ b/common/src/main/java/io/novafoundation/nova/common/data/announcements/AnnouncementMapper.kt
@@ -1,0 +1,26 @@
+package io.novafoundation.nova.common.data.announcements
+
+import io.novafoundation.nova.common.domain.announcements.Announcement
+import io.novafoundation.nova.common.domain.announcements.AnnouncementSection
+
+fun mapAnnouncementsFromRemote(
+    remote: AnnouncementsRemote,
+    section: AnnouncementSection,
+    languageCode: String
+): List<Announcement> {
+    return remote[section.key].orEmpty().mapNotNull { mapAnnouncementFromRemote(it, languageCode) }
+}
+
+private fun mapAnnouncementFromRemote(remote: AnnouncementRemote, languageCode: String): Announcement? {
+    val description = remote.description.localizedOrDefault(languageCode) ?: return null
+
+    return Announcement(
+        style = mapStyleFromRemote(remote.style),
+        description = description
+    )
+}
+
+private fun mapStyleFromRemote(style: String?): Announcement.Style {
+    return Announcement.Style.entries.find { it.name.equals(style, ignoreCase = true) }
+        ?: Announcement.Style.INFO
+}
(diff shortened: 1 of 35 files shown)


common/src/main/java/io/novafoundation/nova/common/data/announcements/AnnouncementMapper.kt
```diff
@@ -0,0 +1,27 @@
...
+fun mapAnnouncementsFromRemote(
+    remote: AnnouncementsRemote,
+    section: AnnouncementSection,
+    languageCode: String
+): List<Announcement> {
+    return remote[section.key].orEmpty().mapNotNull { mapAnnouncementFromRemote(it, languageCode) }
+}
+
+private fun mapAnnouncementFromRemote(remote: AnnouncementRemote, languageCode: String): Announcement? {
+    val description = remote.description.localizedOrDefault(languageCode) ?: return null
```
stepanLav: `description` is declared non-null, but Gson doesn't enforce Kotlin nullability. An entry in nova-utils without `description` (or a `null` element in the section array on line 11) lands here as null and throws NPE.

The network call is guarded by `runCatching` in the repository, but the mapping isn't, so this escapes the `flow {}`. Consumers use `shareInBackground()` = `shareIn(viewModelScope, Eagerly)`, and `BaseViewModel` installs no `CoroutineExceptionHandler` — so it's an uncaught exception and the staking tab crashes. Worst part is it ships from a JSON push, no app release needed.

```kotlin
// AnnouncementsRemote.kt
val description: Map<String, String>?

// here
val description = remote.description?.localizedOrDefault(languageCode) ?: return null
```

Would also add a `.catch { Log.e(...); emit(emptyList()) }` on the repository flow as a backstop, so no future mapping change can take the screen down.

common/src/main/java/io/novafoundation/nova/common/data/announcements/AnnouncementsRepository.kt
```diff
@@ -0,0 +1,47 @@
...
+
+class RealAnnouncementsRepository(
+    private val announcementsApi: AnnouncementsApi,
+    private val contextManager: ContextManager
+) : AnnouncementsRepository {
+
+    @Volatile
+    private var cache: AnnouncementsRemote? = null
+
+    override fun announcementsFlow(section: AnnouncementSection): Flow<List<Announcement>> = flow {
```
stepanLav: This is a cold flow, so every collector refetches the file. `StakingDashboardViewModel` collects it twice (`announcements` and `announcementsByChain` are two independent `shareInBackground` chains), `StakingViewModel.announcementFlow` recollects on every chain switch via `flatMapLatest`, and `StartStakingLandingViewModel` adds another — opening the dashboard alone is 2 GETs of the same JSON.

We already have this exact pattern in `RealGlobalConfigDataSource` (mutex + fetch once, cache). Either reuse that shape or share the flow once in the repository:

```kotlin
private val allAnnouncements = flow { ... }
    .shareIn(appScope, SharingStarted.WhileSubscribed(), replay = 1)
```

Otherwise we end up with a second, weaker mechanism for the same job.

feature-staking-impl/src/main/java/io/novafoundation/nova/feature_staking_impl/presentation/announcements/AnnouncementModel.kt
```diff
@@ -0,0 +1,49 @@
...
+private fun mapAnnouncementStyleToUi(style: Announcement.Style): AlertView.StylePreset {
+    return when (style) {
+        Announcement.Style.INFO -> AlertView.StylePreset.INFO
+        Announcement.Style.WARNING -> AlertView.StylePreset.WARNING
+        Announcement.Style.ERROR -> AlertView.StylePreset.ERROR
+    }
+}
+
+@ColorRes
+private fun mapAnnouncementBackgroundToUi(style: Announcement.Style): Int {
```
stepanLav: These two mappers reproduce `AlertView.Style.fromPreset` verbatim — same colors, same drawables, same null tint. DRY: a palette change now needs edits in both places.

`AlertView.kt` already exposes `fun AlertView.StylePreset.asStyle()`, so the model can keep just `stylePreset` + `description`, and `DashboardHasStakeAdapter.bindAnnouncement` reads `model.stylePreset.asStyle().backgroundColorRes` / `.iconRes`. `AnnouncementsAdapter` doesn't touch these two fields at all, so nothing else is lost.

feature-staking-impl/src/main/java/io/novafoundation/nova/feature_staking_impl/domain/announcements/StakingAnnouncementsUseCase.kt
```diff
@@ -0,0 +1,40 @@
...
+interface StakingAnnouncementsUseCase {
+
+    fun generalAnnouncementsFlow(): Flow<List<Announcement>>
+
+    fun announcementsByChainFlow(): Flow<Map<ChainId, Announcement>>
+
+    fun announcementFlow(chainId: ChainId): Flow<Announcement?>
+}
+
+class RealStakingAnnouncementsUseCase(
```
stepanLav, line 19: Can we get unit tests for the new logic? Everything interesting is a pure function: `localizedOrDefault` falling back to the `default` key, `mapStyleFromRemote` degrading to INFO on an unknown style, and the general-vs-per-chain split here.

`feature-staking-impl/src/test` and `common/src/test` are already set up, and these are exactly the paths a malformed config file hits first — which is also where the NPE above lives.

common/src/main/java/io/novafoundation/nova/common/data/announcements/AnnouncementsRepository.kt
```diff
@@ -0,0 +1,47 @@
...
+import io.novafoundation.nova.common.domain.announcements.Announcement
+import io.novafoundation.nova.common.domain.announcements.AnnouncementSection
+import io.novafoundation.nova.common.resources.ContextManager
+import kotlinx.coroutines.flow.Flow
+import kotlinx.coroutines.flow.distinctUntilChanged
+import kotlinx.coroutines.flow.flow
+
+private const val LOG_TAG = "AnnouncementsRepository"
+
+interface AnnouncementsRepository {
```
stepanLav, line 19: A short KDoc on the new public surface would help here (and on `StakingAnnouncementsUseCase` / `localizedOrDefault`). Three things aren't inferable from the code: what the `default` key inside `description` means, that `chainId == null` is what marks a general announcement, and that an unrecognised style deliberately falls back to INFO rather than being dropped.

stepanLav (approved):

PR #2321, opened 2026-08-27
docs: fix sovereign spelling in account matcher

wangjingshuiku:
Correct “soveregin” to “sovereign” in the comment describing sibling parachain sovereign accounts.

The matcher prefixes and Kotlin behavior are unchanged; this only makes the system-account classification comment accurate and searchable.


wangjingshuiku:
@svojsu @antonijzelinskij Could you please take a look when you have a chance? Thanks!

PR #2333, opened 2026-09-17
10.10.0

antonijzelinskij:


antonijzelinskij:
Release severity: Normal
Release notes:
— Improved asset management: your asset list, by default, now only shows the tokens that matter to you, such as the ones you own - additionally, any supported token can still be added in a few taps
— Fixed an issue where Hydration swaps with a specified receive amount could fail with a slippage error
— Various improvements and stability fixes

svojsu (approved):

